PR #14, opened 2019-04-04
Make sure we have a default widget during field instance setup

DiegoPino:
# FIXES

Main issue here is how Drupal 8 decides which widget to show when setting a field inside the Content Type Admin menu. There you can set default values, etc, but if a special $form_state key is not present *which can be only set via a form alter and is not used anywhere i core* if will render the widget used in the default form display to set The defaults. That is in our case totally wrong, we don't want to set a default JSON via a webform in this context (and it also does not work).

# IMPROVES
Since our StrawberryField Field uses mostly computed Properties i wanted to improve computing times. We now use much more ::computed to check instead of going into ::processed content on each property. There are tiny changes here, but it will drop a few extra cycles if the main property of the field is empty (->value) and will also be way more responsive on calculating if the field is empty which seems something Drupal does a lot.

@giancarlobi could you test this branch in any of your Archipelagos to see if everything keeps working, Solr indexing is still there and also on creating a new Object using the default Widget without any default JSON values, you see no errors?

Thanks 😄

### DIFF
--- a/src/Field/StrawberryFieldItemList.php
+++ b/src/Field/StrawberryFieldItemList.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Drupal\strawberryfield\Field;
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Field\FieldItemList;
+
+
+/**
+ * Represents an entity field; that is, a list of StrawberrField item objects.
+ *
+ * An entity field is a list of field items, each containing a set of
+ * properties. Note that even single-valued entity fields are represented as
+ * list of field items, however for easy access to the contained item the entity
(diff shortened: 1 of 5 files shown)


giancarlobi:
@DiegoPino all works good, also a new content with empty JSON. Great!

DiegoPino:
Hi! With empty JSON you mean {} ? I hope so, if not i made a mistake =) Thanks

giancarlobi:
I mean with an empty default here admin/structure/types/manage/digital_object/fields/node.digital_object.field_descriptive_metadata. Before fix, when select Add node as Super admin Raw JSON (node/add/digital_object/direct_json_editing) gives me an error, now no errors.

PR #19, opened 2019-05-06
Provide Event driven reactions to Strawberry Bearing Content Entities

DiegoPino:
# What is new

To allow for much more control on how the JSON's lifecycle inside a StrawberryField happens, from its conception either via Webforms, direct JSON editing (manual) or API we decided to address some of our many `@TODOS` regarding an event driven architecture and allowed SBF to define events, dispatch them and set the basic abstract classes from which listeners to those events can be extended and built on.

There is a tiny little bit of performance improvement here too, since we cache statically now which Bundles bear Strawberry fields before dispatching the events and pass the machine name of those fields to the event itself.  Function which could be called many times in a batch ingest process will keep its results around during a single process and make also dispatching events much more safer.

# NOTE

Don't merge yet. No need to test until all commits are there. This is the first of a series of commits that will build this pull, all  needed to address correctly https://github.com/esmero/format_strawberryfield/issues/11  and also needed to move some heavy JSON and file entity lifting from https://github.com/esmero/webform_strawberryfield/. Basically we will move some Webform handler logic to an EventSubscriber and deal with marking which JSON keys refer to entities (files, nodes, etc) in a way more generic way, also moving away from storing Drupal IDs  to storing internal links to the actual entities (smart!). This will make our content totally portable between Drupal's and much easier to ingest.

@giancarlobi will let you know when i'm there and ready for testing

### DIFF
--- a/src/Event/StrawberryfieldCrudEvent.php
+++ b/src/Event/StrawberryfieldCrudEvent.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace Drupal\strawberryfield\Event;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\strawberryfield\StrawberryfieldEventType;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Class to contain a SBF bearing entity event.
+ */
+class StrawberryfieldCrudEvent extends Event {
+
+  /**
+   * The Entity.
+   *
+   * @var \Drupal\Core\Entity\EntityInterface
+   */
+  private $entity;
+
+  /**
+   * The SBF field machine names.
+   *
+   * @array;
+   */
+  private $fields;
+
+  /**
+   * The event type.
+   *
+   * @var \Drupal\strawberryfield\StrawberryfieldEventType
+   */
+  private $eventType;
+
+  /**
+   * Construct a new entity event.
+   *
+   * @param string $event_type
+   *   The event type.
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity which caused the event.
+   */
+  public function __construct($event_type, EntityInterface $entity, array $sbfFields) {
+    $this->entity = $entity;
+    $this->eventType = $event_type;
+    $this->fields =  $sbfFields;
+  }
+
+  /**
+   * Method to get the entity from the event.
+   */
+  public function getEntity() {
+    return $this->entity;
+  }
+
+  /**
+   * Method to get the fields from the event.
+   */
+  public function getFields() {
+    return $this->fields;
+  }
+
+  /**
+   * Method to get the event type.
+   */
+  public function getEventType() {
+    return $this->eventType;
+  }
+
+}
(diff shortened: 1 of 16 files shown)


giancarlobi:
@DiegoPino perfect, I'm here ready for testing when you want.

DiegoPino:
@giancarlobi i just added this https://github.com/esmero/strawberryfield/pull/19/commits/6b117584a6bf1034bf254f522243e8a4267aeacc commit and i would like a second set of eyes.
 
One of the SBF needs is to able to allow us, and other systems, to have access to actual JSON in different stages of the process. To Enrich, filter, validate, complement, reconcliate, etc. Many of those needs are absorbed my the CRUD events i committed in this pull too, but, sometimes there is not an actual Entity in place, or because of safety and speed, i just want to pass JSON around and get back a processed JSON. So i added a new Event type there that has two JSON based properties (a raw, reference JSON and one to be processed by the Subscribers). Copying my commit questions here

> I'm having my doubts right now if passing an arrays here instead of JSON encoded Objects are the right way, Arrays are easier to manage to people making changes to the data, but Objects have the benefit that i don't need to write the changes back to Event since PHP keeps a pointer to the original OBJECT.
Sadly if i go JSON ENCODE OBJECT i will force people to decode, mangle and reencode and in that case the reference/pointer is of no use anymore.
Another option would be to create a special Custom Object that has both arrays as elements. That way i can still access the arrays directly and write them back. Feels a bit convolued so i'm going now for just Arrays. @giancarlobi ideas? 

What do you think?

giancarlobi:
Assumptions:
A) some events (i.e. JSONPROCESS, PRESAVE) don't modify original SBF JSON and they only prepare a processed JSON so we can abort any of these stages without need of rollback
B) some events (i.e. SAVE) are the final commit and they update the SBF JSON
C) arrays are better than JSON objects to be managed/modified by code

So, A) event subscribers don't need pointer to SBF JSON object as they don't update it while B) event subscribers need pointer to SBF JSON object.

What about two classes that extend events, one with arrays (original and processed SBF JSON) and the other with a JSON object (pointer to original) + array processed JSON ?

DiegoPino:
@giancarlobi i agree.
A) JSONPROCESS, PRESAVE: 
- JSONPROCESS only works on passed JSON but we make the decision if persisting, merging or discarding what ever comes out of that Subscriber processing chain. 
- PRESAVE: Presave has access to the actual Entity Object,(alias DO or simply NODE) and can add, remove, modify values directly into the Objects's SBF field, but, since the Object is still not persisted to Storage, that is totally OK. We can always stop the chain at the end. We always try to encode the JSON back and only INSERT/SAVE if no JSON Errors are thrown
B) SAVE AND INSERT are final. Yes. Same, we always try to encode the JSON back and only INSERT/SAVE if no JSON Errors are thrown
C) Yes. Way easier for non super-duper-oop devs!
So on A). JSONPROCESS only needs access to an Array, and  on an entity presave hook i would need to dispatch, probably both. One directly with the Entity Attached, and one with just JSON. Reason for this is to avoid duplicating logic (like enrich the JSON) in two different subscribers, since i will be dispatching the JSONPROCESS also on the Webform handler, to allow some of the heavier logic to be processed there, if people use that ingest workflow, but then also allow the same logic to be triggered for API ingest. Does that make sense?

DiegoPino:
@giancarlobi  closing this. All this changes went into 8.x-1.0-beta1 already.

Issue #42, opened 2019-09-04
React to our SBF Entity Delete Event


DiegoPino:
# What?

We dispatch all type of cool events so other modules or this one can react, modify and do things based on what happened, but we have been waiting for stability and me coming back from vacations (and the wait is over!) to actually start removing referenced files (referenced inside a Strawberry field JSON) Drupal tracking. For those who don't know what that is: we track each file's usage inside a JSON via Drupal 8's file tracking capabilities. That way same file can be used in many place (try the "clone" option for and digital object and you will see!) but also nobody can delete them as long as they are in use. (safe, safe). Now we want to remove tracking when a Digital Object gets purged. Once a file is not tracked anymore, the next cron run will get rid of it.

# How?

Simple as always. We extend one of our abstract classes that knows how to react to that delete event, we pass the deleted entity in the event, we check if there are files tracked for it, we remove them, then we do additional cleanup in case someone else is falsely tracking it too (Drupal's file managing is quite basic and prone to failure so we have to be double strict and double safe)  and if there are false positives, we also remove those. 

Pull coming.

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/e1b01cc4d5b0638a28858962819b62bc11423ad0
Thanks @giancarlobi for your review.

PR #43, opened 2019-09-04
ISSUE-42: React to SBF Digital Object delete event + improvements.

DiegoPino:
# What is new?

See #42 

The logic here is not too complex but requires some explaining:
1. We add a new Subscriber, `StrawberryfieldEventDeleteFileUsageDeleter`, that extends the entity delete abstract class we already have.
2. This even subscribers calls a new method in our `strawberryfilepersister` service, `$this->strawberryfilepersister->removeUsageFilesInJson($entity);`
3. New method checks all files being currently tracked by this entity.
4. Removes all tracking set to this entity.
5. Checks if the file is still tracked by someone else.
6. If not, sets it to temporary so it can be removed by cron (this differs a bit form D8.4+ since that is a setting people can set, but also tricky since there is no mechanist to allow an user/admin to remove a file...so we assume we need to do this). Note: Drupal's file managment is so fragile and naive..
7. If still tracked, we check if the tracked entities really exist! (because of many D8 bugs and failures they could be gone) and we clean up removing broken records only for files tracked originally for the deleted entity (See this as a as-you-go clean up)

@todo. We need a setting that allows files to be kept as "Preservation copies" and tracked outside the DO/Node realm, so our deposited DO as files can keep their references around if we remove the node. Or said in other ways, a check that says "preserve things in storage even if removed from Drupal". Good first timer issue to build a form and a config setting for that we can react on.

# Improvements

This pull adds a new event dispatcher. This dispatcher, `StrawberryfieldEventType::JSONPROCESS`will pass a file $info array structure around, the one that is generated by us via our file router and that has md5, etc for each referenced binary. Why? We want to enrich that via other modules, like counting PDF pages, etc. We don't implement the Subscriber for that here, since SBF is already too heavy and that is why we have now (or will have once i push) https://github.com/esmero/strawberry_runners (yay!) as a not-a-microservice module with `reactphp` capabilities. So cool.

# Fixes

See https://github.com/esmero/strawberryfield/compare/8.x-1.0-beta1...ISSUE-42?expand=1#diff-376f27afff874dbd089c366d26b52febR212
File naming cleanup seems to be a bit off, and it seems that even if the resulting file names are OK for Drupal's use, Cantaloupe fails to resolve things like 
`albany--boats-on-the-hudson-river-(the-jacob-h.-tremper-and-the-fort-grange,-july-1926.-en-image-18c3f04c63b3a16b29fd11ae5bd0386b28e9219a.jp2`

I have not tested this last one yet (blind fix... bad idea when dealing with Regular expressions) so could end adding an extra commit tonight.

Anyway, all needs some testing, my internal pass proved to be successful but i don't trust myself right now =)

@giancarlobi for your eyes here.


giancarlobi:
@DiegoPino looking at code I think it's good, not tested.


> See https://github.com/esmero/strawberryfield/compare/8.x-1.0-beta1...ISSUE-42?expand=1#diff-376f27afff874dbd089c366d26b52febR212
> File naming cleanup seems to be a bit off, and it seems that even if the resulting file names are OK for Drupal's use, Cantaloupe fails to resolve things like
> `albany--boats-on-the-hudson-river-(the-jacob-h.-tremper-and-the-fort-grange,-july-1926.-en-image-18c3f04c63b3a16b29fd11ae5bd0386b28e9219a.jp2`


About this, I executed the command and result is without any change:
```
php > echo preg_replace('([^\w\s\d\-_~,;:\[\]\(\)\].]|[\.]{2,})','', "albany--boats-on-the-hudson-river-(the-jacob-h.-tremper-and-the-fort-grange,-july-1926.-en-image-18c3f04c63b3a16b29fd11ae5bd0386b28e9219a.jp2");
albany--boats-on-the-hudson-river-(the-jacob-h.-tremper-and-the-fort-grange,-july-1926.-en-image-18c3f04c63b3a16b29fd11ae5bd0386b28e9219a.jp2
```
Is it right? what is the expected change to string?

DiegoPino:
@giancarlobi thanks. Nope, that is not right, my regex or the logic is off there. I will give that a few spins in the morning and come back with a better approach and will let you know.

DiegoPino:
@giancarlobi i think i addressed the Naming issue correctly now. Could you give this a look when you have some time?

basically it does this (most of this was there already but now a little bit nicer

- Adds the first part of the mime type as prefix
- Calls new sanitizeFileName
- lower cases
- Deals with UTF8 to ASCII mapping (just to be safe)
- replaces all non known UTF8 chars with -
- replaces all dangerous characters with -
- de duplicates '-'
- trims any '-'
- cuts length to allow fully S3 compliant name (<= than 512 characters). Also good for non S3 users.
- appends the uuid of the file at the end.

let me know what you think!

src/StrawberryfieldFilePersisterService.php
```diff
@@ -191,39 +193,51 @@ public function getDestinationUri(
     // If no destination scheme was setup on our global config use the original file scheme.
     $desired_scheme = !empty($this->destinationScheme) ? $this->destinationScheme : $file_parts['destination_scheme'];
 
-    $destination_filename = $file_parts['destination_filetype'] . '-' . $file_parts['destination_filename'] . '.' . $file_parts['destination_extension'];
+    // Mime type becomes suffix. Performant for filtering in S3.
```
giancarlobi: @DiegoPino I think the comment is wrong (code is OK): 
// Mime type becomes prefix. UUID becomes suffix. Performant for filtering in S3.

src/StrawberryfieldFilePersisterService.php
```diff
@@ -191,39 +193,51 @@ public function getDestinationUri(
...
-      $destination_basename
-    );
-    $destination_basename = preg_replace('/\s+/', '-', $destination_basename);
-    $destination_basename = trim($destination_basename, '-');
-    // If the basename if empty use the element's key.
+    // Edge case, should rarely happen.
     if (empty($destination_basename)) {
-      $destination_basename = $file_parts['destination_filetype'] . '-' . $file->uuid(
-        );
+      $destination_basename = $file_parts['destination_filetype'] . '-unnamed';
```
giancarlobi: @DiegoPino If empty $destination_basename it means $file_parts['destination_filetype'] id empty or empty after sanitize so I suggest :
`$destination_basename = 'nomimetype' . '-unnamed';`

DiegoPino: HI, yep, you are right. My logic there was so wrong, mind still not totally active. Fixing now.
I will go for 'application' instead of 'nomimetype' just to conform with how we calculate that part when not empty which is first part of the mime type and since unknown mime-type is always application/octet-stream we end having application. That if is really an overkill, since it assuming that Drupal would end having a null in the mime-type and the filename would have only invalid characters, which would all happen at the same time if something goes terrible wrong. But i like to be strict on this since we are dealing with valuable repository items.

DiegoPino: @giancarlobi this should be ready. If you have some time tomorrow to give this a look it would be great! Thanks so much for your feedback =)

giancarlobi: It looks good, great.

Issue #47, opened 2019-10-29
Entity UUID Converter for Routes that start with /do/


DiegoPino:
# What?

All our ADOs are using yoursite.edu/do/uuid as canonical path to access instead of the non sense /node/1 thing. Right now that is being done via Aliases. We need and we want UUIDs , right now it seems easier to have path alias programmed to do that. We do it and it works. But! Path alias does not expand, existing aliased paths for subpaths.

So we need a final solution (never final i know) that deals with any Content Entity under the /do idea because things like /do/uuid/metadata/iiifmanifest does not resolved automatically. Since our metadata exposed endpoints are dynamic (good!) we want to access them also via the uuid path.

I opened this originally inside here https://github.com/esmero/format_strawberryfield/issues/25
but now i'm clear the right way of doing this is to create a Resolver Class/event subscriber that is, based on certain conditions, upcast UUIDS and convert them into Loaded Entities. Something in the tone of https://git.drupalcode.org/project/jsonapi/blob/8.x-2.x/src/ParamConverter/EntityUuidConverter.php but simpler, just for us.

Working on that. Will take me more days but on the right track.

DiegoPino:
So i have code and it works!!!! (also it loads the latest revision and works (hopefully, my nightmare) in a language aware context. See https://github.com/esmero/strawberryfield/commit/860d473b8d28e1beefba7e3bdda5459a6f83e144
I will make a pull tomorrow and will also update format_strawberryfield so testing is easier.

DiegoPino:
Fixed via https://github.com/esmero/strawberryfield/commit/859c8c28f83764f87be97d4118f7dfd1dd5724f5

PR #48, opened 2019-10-31
ISSUE-47: UUID node up-caster for Routes that have as parameter 'ado' as resource_type

DiegoPino:
# What is new?

See #47 

To see this in action, momentary (until we merge all). fetch and checkout this branch https://github.com/esmero/strawberryfield/tree/ISSUE-47

in the format_strawberryfield module

either:
A)
edit `format_strawberryfield.routing.yml` and replace 

format_strawberryfield.metadatadisplay_caster: with
```YAML
format_strawberryfield.metadatadisplay_caster:
  path: '/do/{node}/metadata/{metadataexposeconfig_entity}/{format}'
  methods: [GET]
  defaults:
    _controller: '\Drupal\format_strawberryfield\Controller\MetadataExposeDisplayController::castViaTwig'
  options:
    parameters:
      node:
        type: 'entity:node'
      metadataexposeconfig_entity:
        type: 'entity:metadataexpose_entity'
      resource_type:
        type: 'ado'
  requirements:
    format: .+
    _entity_access: 'node.view'
```

or B)

Test with https://github.com/esmero/format_strawberryfield/pull/32

If you have an active Metadata exposed endpoint, like i have at
http://localhost:8001/do/394dbfb7-3676-4963-a4e0-4adef7c84dfa/metadata/iiifmanifest/manifest.jsonld

clear caches first. docker exec -ti esmero-php bash -c "drush cr"
Then navigate to any Digital Object and add to the url /metadata/nameofyourendopint/manifest.jsonld (if the settings match). 

Node should load correctly

This is needed and does not affect anything else  (thanks to `public function applies($definition, $name, Route $route) {`)

@marlo-longley ping

### DIFF
--- a/src/ParamConverter/UuidEntityConverter.php
+++ b/src/ParamConverter/UuidEntityConverter.php
@@ -1,0 +1,104 @@
+<?php
+/**
+ * @file
+ * Contains UuidEntityConverter.php
+ *
+ * @author Diego Pino Navarro <dpino@metro.org> https://github.com/diegopino
+ */
+
+namespace Drupal\strawberryfield\ParamConverter;
+
+
+use Drupal\Core\ParamConverter\EntityConverter;
+use Symfony\Component\Routing\Route;
+use Drupal\Core\Plugin\Context\Context;
+use Drupal\Core\Plugin\Context\ContextDefinition;
+use Drupal\Core\Entity\EntityRepositoryInterface;
+
+/**
+ * Converts an UUID into a valid Content Entity.
+ *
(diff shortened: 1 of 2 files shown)


DiegoPino:
@marlo-longley thanks! The implications are many but for now the most important piece is that on the API side of things,  it also allows us to use the canonical URL for Nodes bearing a SBFs `/do/uuid` we use to access a Node as a rendered piece as base  to add other services as subroutes. It feels obvious, but drupal does not it be default in core (but in the wishlist for 9)

Issue #46, opened 2019-10-18
New JSON Key providers needed


DiegoPino:
# What is needed?

As we move forward and our SBF JSON becomes more rich, we should start thinking and coding new types of JSON Key Provider Plugin implementations. This is related to #33 but goes beyond.

The ones i want and need:
 

1. An Entity Reference Property. This is extremely useful to Index in Solr other nodes and entities referenced from inside the JSON to allow Drupal to see them natively. Specially since our JSON graphs are directed, means many times Parents will point to Children, but also on Collection Membership, we want Collection Descriptions to lead, on search, to children.

**The implementation is quite simple:**

ContentEntityBase provides already a method:

\Drupal\Core\Entity\ContentEntityBase::referencedEntities

Which goes field by field checking for EntityReference Properties
```PHP
 /**
   * {@inheritdoc}
   */
  public function referencedEntities() {
    $referenced_entities = [];

    // Gather a list of referenced entities.
    foreach ($this->getFields() as $field_items) {
      foreach ($field_items as $field_item) {
        // Loop over all properties of a field item.
        foreach ($field_item->getProperties(TRUE) as $property) {
          if ($property instanceof EntityReference && $entity = $property->getValue()) {
            $referenced_entities[] = $entity;
          }
        }
      }
    }

    return $referenced_entities;
  }
```

So what we need is a JSON key provider that exposes a set (one or more) JSON property values (node ids for example ) as  \EntityReference class properties. 

We have at least two ways of providing the JSON keys as arguments:

First, automatic, by using the new  "ap:entitymapping": [] key we preprocess (or should because webform maintainer dismissed my pull request for that...gosh)

Or by allowing people simply to type the keys (hopefully in this case a full JSON Path?) that contain entity references. Example are ismemberof, scene, etc.
With that Solr will allow us to co-index those referenced entities values, like their labels, etc.

2. Make use of our SBF Vocabulary generator to automatically expose all Keys we are SURE will exist across the repository. Basically, if you go to http://localhost:8001/admin/structure/taxonomy/manage/strawberryfield_voc_id/overview
All those vocabularies were generated by content indexed. So, we should allow people to use that directly too, instead of choosing, adding manually keys.

**Here is how i envision that:**

Only keys that should be exposed are the leafs of a branch.

So if we have :
- as:document.*
  - checksum 
  - crypHashFunc 
  - dr:fid 
  - dr:for 
  - dr:uuid

[...]

PR #49, opened 2019-12-03
ISSUE-46: Provide a JMESPath option to expose parsed results as a Field property

DiegoPino:
# What is new?

See #46. This is just one of the listed there. But we are moving forward.

This is just good, lovely and needed. 
This pull provides the following new options:

1. A new JSON Key Name Provider Plugin that deals with JMESPath processing to populate a given Strawberryfield exposed property (For Solr). This Plugin allows a comma separated string of JMESPaths, which means we can take a few different values from all over the JSON and expose them as a single Solr Field/Facet. This also gives us the change to target pretty specific values in a complex hierarchical JSON.

2. A New `DataType` extended from `itemList` matching the new Plugin. It does some smart things like checking if the JMESPath results are actually just an array of values or a Multidimensional Array. In the later case, it JSONEncodes the values and gives a string back. If that is the case, it is really not useful for faceting, true, but yes for free search.

3. New helper methods on our `StrawberryfieldJsonHelper`class including
`::arrayIsMultiSimple(array $sourcearray =  [])`
Simplistic checker for Multidimensional Arrays. Not so fast but pretty, good for small arrays like what we would get from a SBF evaluated against a JMESPath
`::arrayIsMultiIterative(array $sourcearray =  [])`
Pretty similiar (same input/output) but uses a foreach. Which for larger arrays is way faster. Probably useful when working on Find and Replace and larger serializations.
`::jsonPropertyPathToJmesPath($jsonproppath)`
A function that converts JSONPath syntax into a valid JMESPath syntax. Deals with simple use cases, not complex casting of results, but will be used when parsing our existing Vocabularies/Paths into valid JMESPaths

# How to test?

- Checkout this Branch
- Clear Caches
- Go to `/admin/structure/strawberry_keynameprovider`
- Press `Add Key Name Provider`
- Give it a Label: `Subjects`
- Select the new Plugin from "Strawberry Key Name Provider Plugin" named JmesPath Strawberry Field Key Name provider. Ajax will give you more options down there.
- Under `One or more comma separated valid JMESPaths.` write the following:
`subject_loc[*].label, subject_wikidata[*].label`
Which in JMESPath logic means, take all `label` values in any List item under `subject_loc` JSON key, and take all `label` values  in any List item under `subject_wikidata` JSON Key
Under the Input you will see a Machine name (Exposed Strawberry Field Property: subject_loc_uri). Press `Edit`and change it to `all_subjects`.

Press `Save`. Clear Caches (just in case)

Now go to your Search Index and add a new Field from Descriptive Metadata. Search for the new `all_subjects` property.
You can also add this one (so you see this is working!) as a new Facet, and expose the Facet block.

Now you will have a new facet block that joins all the labels coming from Wikidata and LoC Subjects in the same place. Pretty handy!

# Other Uses
Simple stuff. Like just expose the Primary `type` key instead of all the `type` keys (guess what!, the JMESPath for that is `type` jajaja). Also, if JMespath is as simply i could add logic to avoid doing a JMEspath query at all. Internal `@TODO`, no need to expose people to the choice.

# What is next

- Allow people to select the JMESPaths from our existing Vocab using autocomplete, since we are already keeping track of all Paths on Save. That would lover the barrier.

- Add some smarter things here. I wish in the future we had less Cache clearing, which is really something we can do by triggering that on save.

- Allow people to evaluate the JMesPaths agains a choosen ADO. That way they can see results before committing to a given Query. Right now you can also do some test on a terminal via JPTerm. Pretty handy for learning JMESPath. See https://github.com/jmespath/jmespath.terminal

- Add more validation and checks. This is really missing. Specially on that Settings form.

Thoughts? @giancarlobi @marlo-longley 
Thanks!

### DIFF
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -44,6 +44,17 @@ strawberryfield.strawberry_keynameprovider.flavor:
       type: string
       label: 'The field property we expose for the Strawberryfield'
 
+strawberryfield.strawberry_keynameprovider.jmespath:
+  type: config_object
+  label: 'Strawberry Key Name Provider JMesPath specific config'
+  mapping:
+    source_key:
+      type: string
+      label: 'A Comma separated string containing one or more JMESPaths '
+    exposed_key:
+      type: string
+      label: 'The field property we expose for the Strawberryfield'
+
 strawberryfield.storage_settings:
   type: config_object
   label: 'Strawberry Storage specific config'
--- a/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
+++ b/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
@@ -1,0 +1,187 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: dpino
+ * Date: 9/18/18
+ * Time: 8:21 PM
+ */
+namespace Drupal\strawberryfield\Plugin\DataType;
+use Drupal\Core\TypedData\Plugin\DataType\ItemList;
+use Drupal\strawberryfield\Plugin\Field\FieldType\StrawberryFieldItem;
+use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
+
+class StrawberryValuesViaJmesPathFromJson extends ItemList {
+  /**
+   * Cached processed value.
+   *
+   * @var array|null
+   */
+  protected $processed = NULL;
+  /**
+   * Whether the values have already been computed or not.
+   *
+   * @var bool
+   */
+  protected $computed = FALSE;
+  /**
+   * Keyed array of items.
+   *
+   * @var \Drupal\Core\TypedData\TypedDataInterface[]
+   */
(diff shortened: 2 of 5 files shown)


DiegoPino:
@mitchellkeaney testing Slack integration

giancarlobi:
@DiegoPino Great!! I'll test it on my archipelago w/o docker. (Yeah, I come back from last "trojan invasion" ... ).

DiegoPino:
@giancarlobi thanks! I was avoiding bothering you as much as possible, i know you are super busy, so really double thank you.

giancarlobi:
@DiegoPino At the first try .. it worked great!!! I'll check a little bit deeper tomorrow.

image

DiegoPino:
You are a rocket! Thanks! Looking good. Will give this a few more days so you can also give me your feedback on the code (while i figure out a way of upgrading to 8.8.0 that does not break everything jajajaj).

DiegoPino:
@giancarlobi @marlo-longley moving forward with this. Merging!

Issue #53, opened 2020-01-07
Create config for the Solr field used in Type/View Mode mapping 


marlo-longley:
### What is needed?

Right now we can configure ADO by Type to map onto View Modes in Drupal.
See: `/admin/config/archipelago/viewmode_mapping` in an Archipelago instance.

Solr field for Type is fixed/hardcode currently.
See [here](https://github.com/esmero/format_strawberryfield/blob/8.x-1.0-beta1/src/Form/ViewModeMappingSettingsForm.php#L226). 

Tasks:
1. Make a config form that allows user to set which Solr Server, Index, and then Field will be used in the ADO Type/View Mode mapping. 
2. And then change the `ViewModeMappingSettingsForm` to draw from this config, rather than hardcode.

#### Notes
I need to think about naming for this. It's kind of clunky. I started using "Type to Solr Field" for the name of the form which probably isn't clear.

DiegoPino:
Partially Solved via https://github.com/esmero/strawberryfield/commit/30fd84f3541dc52cf02cb6165511266dd25d853c
Will do the format_strawberryfield pull now to close the circle

aksm:
@DiegoPino was this circle ever closed?

DiegoPino:
Please check the source code

DiegoPino:
Fixed

PR #54, opened 2020-01-08
ISSUE-53: Create new Config page for Solr settings

marlo-longley:
Addresses #53 

### Issue Summary
This push addresses the first part only of #53, creating the form to save the new config.

### What's changed?
- Add schema for strawberryfield.archipelago_solr_settings. For now it only holds type related config
- Add route to /admin/config/archipelago/important-solr-settings
- Add ImportantSolrSettingsForm.php
  This is the most code in this file. I don't know if it is the optimal UI, with multiple dropdowns, but it works pretty well for me.

### To Test
Go to /admin/config/archipelago/important-solr-settings and follow the form
I created extra Servers and Indexes to make sure all the conditional switching worked.

### Notes
Not sure how you want to deal with no config set. I am guessing, @DiegoPino since you were going to deal with the second part of #53 , that you might set default config.

### DIFF
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -66,20 +66,6 @@ strawberryfield.storage_settings:
       type: string
       label: 'Storage Scheme for Persisting Digital Objects in JSON format'
       
-xstrawberryfield.archipelago_solr_settings:
-  type: config_object
-  label: 'Important Solr settings for Archipelago'
-  mapping:
-    type_server: 
-      type: string
-      label: 'The Solr server for type in ADO to View Mode mapping'
-    type_index:
-      type: string
-      label: 'The Solr index used for type in ADO to View Mode mapping'
-    type_field:
-      type: string
-      label: 'The Solr field used for type in ADO to View Mode mapping'
-      
 strawberryfield.archipelago_solr_settings:
   type: config_object
   label: 'Important Solr Settings for Archipelago'
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -65,3 +65,17 @@ strawberryfield.storage_settings:
     object_file_scheme:
       type: string
       label: 'Storage Scheme for Persisting Digital Objects in JSON format'
+      
+strawberryfield.archipelago_solr_settings:
+  type: config_object
+  label: 'Important Solr settings for Archipelago'
+  mapping:
+    type_server: 
+      type: string
+      label: 'The Solr server for type in ADO to View Mode mapping'
+    type_index:
+      type: string
+      label: 'The Solr index used for type in ADO to View Mode mapping'
+    type_field:
+      type: string
+      label: 'The Solr field used for type in ADO to View Mode mapping'
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -66,7 +66,7 @@ strawberryfield.storage_settings:
       type: string
       label: 'Storage Scheme for Persisting Digital Objects in JSON format'
       
-strawberryfield.archipelago_solr_settings:
+xstrawberryfield.archipelago_solr_settings:
   type: config_object
   label: 'Important Solr settings for Archipelago'
   mapping:
@@ -79,3 +79,23 @@ strawberryfield.archipelago_solr_settings:
     type_field:
       type: string
       label: 'The Solr field used for type in ADO to View Mode mapping'
+      
+strawberryfield.archipelago_solr_settings:
+  type: config_object
+  label: 'Important Solr Settings for Archipelago'
+  mapping:
+    ado_type:
+      type: strawberryfield.archipelago_solr_settings.field_config
+      label: 'Source for Archipelago Digital Object Type'
+      
+strawberryfield.archipelago_solr_settings.field_config:
+  type: mapping
+  label: 'Field Config for Archipelago Solr Settings'
+  mapping:
+    index_id:
+      type: string
+      label: 'ID of Solr Index'
+    field:
+      type: string
+      label: 'Solr Field'
+      
--- a/src/Form/ImportantSolrSettingsForm.php
+++ b/src/Form/ImportantSolrSettingsForm.php
@@ -30,33 +30,38 @@ class ImportantSolrSettingsForm extends ConfigFormBase {
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
   protected $sbfUtiliy;
-  
+
   protected $solrServers;
-  
+
   /**
    * Constructs an ImportantSolrSettingsForm object.
    *
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
    *   The entity type manager.
-   * 
+   *
    * * @param \Drupal\strawberryfield\StrawberryfieldUtilityService $sbf_utility
    *   SBF Utility Service
-   * 
+   *
    * * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
    *   The factory for configuration objects.
    */
-  public function __construct(EntityTypeManagerInterface $entity_type_manager, StrawberryfieldUtilityService $sbf_utility, ConfigFactoryInterface $config_factory) {
+  public function __construct(
+    EntityTypeManagerInterface $entity_type_manager,
+    StrawberryfieldUtilityService $sbf_utility,
+    ConfigFactoryInterface $config_factory
+  ) {
     $this->entityTypeManager = $entity_type_manager;
     $this->sbfUtiliy = $sbf_utility;
     $this->setConfigFactory($config_factory);
-    $this->solrServers = $entity_type_manager->getStorage('search_api_server')->loadMultiple();
+    $this->solrServers = $entity_type_manager->getStorage('search_api_server')
+      ->loadMultiple();
   }
 
   /**
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container) {
-    
+
     return new static(
       $container->get('entity_type.manager'),
       $container->get('strawberryfield.utility'),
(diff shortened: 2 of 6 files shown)


config/schema/strawberryfield.schema.yml
```diff
@@ -64,4 +64,18 @@ strawberryfield.storage_settings:
...
       type: string
-      label: 'Storage Scheme for Persisting Digital Objects in JSON format'
\ No newline at end of file
+      label: 'Storage Scheme for Persisting Digital Objects in JSON format'
+      
+strawberryfield.archipelago_solr_settings:
+  type: config_object
+  label: 'Important Solr settings for Archipelago'
+  mapping:
+    type_server:
```
DiegoPino: the naming of this properties is not totally accurate and we can make this better. Idea. For each setting, in this case the ADO type, we want them to have their own config map: So in this case the mapping would be, property name "ado_type", type "strawberryfield.archipelago_solr_settings.fieldconfig" and this `fieldconfig`type is a map that has the following properties:
- server: (read further down not needed really)
- index_id:
- field:

Regarding the "server" key. Indexes are entities. That means, that their names are unique, means also, the "server" here is more a commodity to allow people to fetch, via the UI, a given Server, but there is no need  to save it in the configuration, can be extracted from the same index. See https://github.com/esmero/format_strawberryfield/blob/8.x-1.0-beta1/src/Form/ViewModeMappingSettingsForm.php#L220, we never call the server directly, we use the index, fetch the server from it. Same can be done in the form here.

That way we can reuse that structure for other Solr fields we will want to allow people to configure.

src/Form/ImportantSolrSettingsForm.php
```diff
@@ -0,0 +1,249 @@
...
+    
+    $form['type']['solr_field_select'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Step 3. Select Solr Field'),
+      '#prefix' => '<div id="solr-field-select">',
+      '#suffix' => '</div>',
+      '#options' => $config->get('type_server') && $config->get('type_index') ? $this->getSolrFields($config->get('type_index')) : null,
+      "#empty_value" => "",
+      "#default_value" => $config->get('type_field'),
+      // If not #validated, dynamically populated dropdowns don't work.
```
DiegoPino: true

src/Form/ImportantSolrSettingsForm.php
```diff
@@ -0,0 +1,249 @@
...
+      '#default_value' => $config->get('type_server'),
+      '#ajax' => [
+        'callback' => '::onServerSelect',
+        'disable-refocus' => FALSE,
+        'event' => 'change',
+        'wrapper' => 'index-select', 
+      ]
+    ];
+    
+    $form['type']['index_select'] = [
```
DiegoPino: In Step2 and Step3 you can use #states too. When i said #states where not the right way i meant for populating the values. The way you do it via Ajax is the right now, but still, it would be great that in case of no values of Step1, step2 would be not shown, same goes with step3 if step2 is empty.

src/Form/ImportantSolrSettingsForm.php
```diff
@@ -0,0 +1,249 @@
...
+        'wrapper' => 'solr-field-select', 
+      ]
+    ];
+    
+    $form['type']['solr_field_select'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Step 3. Select Solr Field'),
+      '#prefix' => '<div id="solr-field-select">',
+      '#suffix' => '</div>',
+      '#options' => $config->get('type_server') && $config->get('type_index') ? $this->getSolrFields($config->get('type_index')) : null,
```
DiegoPino: We need better logic here, and that can go into its own function:
The Solr field should have a set of certain conditions to be usable (selectable) in this context.

```YAML
 type_1:
    label: 'Descriptive Metadata » type'
    datasource_id: 'entity:node'
    property_path: 'field_descriptive_metadata:type'
    type: string
    dependencies:
      config:
        - field.storage.node.field_descriptive_metadata
```

So in specific, fields that can be used are the ones where their data sources is a node and the property path contains a field that is of type SBF. I know this adds some overhead to your code, but it you can add/reuse some of the functionality we already have in this module, you will notice its not that complex. And we can reuse that logic quite a lot if you generalize it (can go, e.g into the utils service we provide, just saying).
Always check existing configurations so you can see how things are being done (and what data you have available to make smarter selections/filters), e.g the Search Index config YAML
single export

src/Form/ImportantSolrSettingsForm.php
```diff
@@ -0,0 +1,249 @@
...
+    
+    // Get all Solr servers
+    $servers = $this->entityTypeManager
+      ->getStorage('search_api_server')
+      ->loadMultiple();
+
+    // Create a fieldset for the Type setting
+    $form['type'] = array(
+      '#type' => 'fieldset',
+      '#title' => $this->t('Solr Field for Type in View Mode mapping'),
```
DiegoPino: This wording is not clear to me. View Mode Mapping is not defined in this module. I would say this needs to state which Solr field is going to be used as Source for an Archipelago Object's Type.

DiegoPino:
@giancarlobi i added some review notes here for @marlo-longley. If you happen to have some time, could you give them a look? Maybe i missed something. Thanks!

giancarlobi:
@DiegoPino Sure, next days, tomorrow a full day (as today) but Friday could be.

DiegoPino:
@giancarlobi thanks!

src/Form/ImportantSolrSettingsForm.php
marlo-longley: @DiegoPino Thanks. I thinking I needed some clarification on the larger picture. That should help with other naming/labels.

marlo-longley: What do you think of this?

SOURCE FOR ARCHIPELAGO DIGITAL OBJECT'S TYPE

Choose the Solr field that determines an ADO's Type. Type is then used across Archipelago, for example in our ADO to View Mode mapper.

DiegoPino: SOURCE FOR ARCHIPELAGO DIGITAL OBJECT'S TYPE : yes!, but it needs to say also "Field" somewhere (in the description works too). And now that you mention it, the description could 😭 use a link to that index' Field config page, that way, if the field they search/want is not there, with a click they can go and add it. Also, yes, Type is then used across Archipelago, for example in our ADO to View Mode mapper, but that makes only sense if formatter_strawberryfield is enabled (i mean, why would they not have it enabled? but then people can do whatever they want eventually) so i would, maybe? remove the "for example". formatter_strawberryfield could form alter this one and add it, if enabled? What do you think? I don't want to introduce "soft" dependencies between modules.

src/StrawberryfieldUtilityService.php
```diff
@@ -82,4 +82,29 @@ public function bearsStrawberryfield(ContentEntityInterface $entity) {
...
+   * @return array
+   *  Returns array of solr fields
+   */
+  public function getStrawberryfieldSolrFields(Index $index_entity) {
+    $solr_fields = $index_entity->get('field_settings');
+    $sbf_solr_fields = array();
+
+    foreach ($solr_fields as $field) {
+      $property_path = explode(":", $field['property_path']);
+      $has_sbf_property_path = in_array('field_descriptive_metadata', $property_path);
```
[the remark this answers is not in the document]
DiegoPino: Or, you can get all SBF's machine names in the system and use that array of names to check against. Just another option

DiegoPino:
@marlo-longley i will write up here all the issues (so many.. 😢 ) i have found:Please don't react to fixing/re factoring this yet, there are core issues with your code i want to go step by step with you.

1.- Starting from fresh (no settings saved)
`The website encountered an unexpected error. Please try again later.
TypeError: Argument 1 passed to Drupal\strawberryfield\StrawberryfieldUtilityService::getStrawberryfieldSolrFields() must be an instance of Drupal\search_api\Entity\Index, null given, called in /var/www/html/web/modules/contrib/strawberryfield/src/Form/ImportantSolrSettingsForm.php on line 161 in Drupal\strawberryfield\StrawberryfieldUtilityService->getStrawberryfieldSolrFields() (line 135 of modules/contrib/strawberryfield/src/StrawberryfieldUtilityService.php).` OK. That was bad.

Looking deeper at your code and after fixing that initial issue i found that i could not reach step 3, really not even step 2 was Ok (i could not deselect (no --select Solr index-- empty option), e.g the default index, and without the option of doing so, the 'change' even could not be triggered.

2.- You are trying to "build" element options and build render arrays in the Ajax Callback. nope. Its not the way, and it is not way this works. There are infinite examples googling and in our core around. When you do that, the Form registry gets crazy, Drupal looses complete knowledge of what is happening because you are basically sending new elements via AJAX (after the submit happened and the form was already build) to the browser, but the form, the stored cached version, differs already! On the next form rebuild all your changes, which are just "visual/client side" get lost. There are a few exceptions where you want to change things directly, like changing e.g an #attribute, but with the danger that the form could get out of sync and appropriate measures need to be taken. Look for working examples, inclusive in the Drupal Examples module. I guess i will have to give you more Forms to work on 😺 

3.- Your logic is wrong in many parts. E.g you are loading config defaults over and over each time you rebuild, so the passed values via the form state get lost and are useless.

4.- Your Field options (last select box) are an numerically indexed array. So you are storing the position of the Field, not the machine name of the field, totally not what we want, and not really useful.

5.- Form can be saved with invalid values.

**This is what we have to learn from this:**

All logic goes into the Form builder method. That is the brain. On each Rebuild, the form needs to be able to generate a new one, based on the passed values. Imagine if you had no Javascript around. The form, based on your submitted values, needs to be able to render the whole thing. All #states and javascript are just UI experience things. But Code itself needs to feed/react/and work on a super simple promise: 
- We try to load saved values. 
- Once user starts to interact, all is based on the submitted values and the way the form rebuilds itself, the logic of getting the indexes, the fields, etc, based on those values. This happens over and over.
- Once all is done, and the user saves, we stop rebuilding the form. We persist

I will finish cleaning the code now and will push a branch you can test. Not sure if i will be able to solve whole thing in a super clean right now, but at least it should be working. Will let you know when that is done.

DiegoPino:
@marlo-longley https://github.com/esmero/strawberryfield/pull/54/commits/d8f22a2b617189910890c89a7ae1cbca5dea658e abd https://github.com/esmero/strawberryfield/pull/54/commits/46974512616474fc9fc0b0a1324b229450909144 explain all the changes i did. Please review in detail (like read the whole thing and go step by step) and test/debug

There could be still edge cases, but saving, starting from scratch, deciding on a new index and new server once you are on the third step, etc, is working.

Clear your local settings first
`docker exec -ti esmero-php bash -c 'drush cdel strawberryfield.archipelago_solr_settings.ado_type'`

marlo-longley:
@DiegoPino I really appreciate your help. It makes more sense. I will also study this more this week. Need to be better at forms definitely.

As for testing, I pulled the new commits and ran the drush command. All works well.

The thing that tripped me up most was, trying to fully replace the "Step 3." select dropdown with that error message. This was probably a case of me overthinking the UI -- you just added it below the empty dropdown but still visible, and it works just fine. But clearly also, I need to learn forms better.

Thank you, we can discuss more in person.

DiegoPino:
@marlo-longley, yeah, lets discuss this in today's dev call. I get the UI issue, but even so, the logic was not ready. We need to do better and faster. Will accept my own code merge this during the day. Not what i normally like to do but needed here.

PR #63, opened 2020-01-29
ISSUE-61: Make orderSequence a Static method of StrawberryfieldJsonHelper

DiegoPino:
# Whats new? 

Move Array ordering/sorting out of formatter into a static method in our JSON helper class.

https://github.com/esmero/format_strawberryfield/pull/41 introduced a bug (because not every formatter extends our base formatter and i'm sleeping) so a good opportunity to also make that one better by moving this function totally out of our formatter.

Been giving the thought that the SBF field itself could expose a computed property that orders and decodes the JSON string into an array, maybe for Beta3


DiegoPino:
@marlo-longley thanks! Merging the shame away....

DiegoPino:
I got the branch name wrong. But this solves https://github.com/esmero/strawberryfield/issues/64

Issue #74, opened 2020-02-21
Theme hints for ADOS (do/) and JSON key types


DiegoPino:
# What is needed?

The default Drupal 8 theme hinting for node is kinda silly. it uses page-- or page--node some number  like `page--node--12.html.twig`. We really think its cool that we are actually using NODES as ADOS and that we only need one Content Type (Digital Object) but also we want to allow site builders and theming folks to target pages via templates grouped by JSON key type (book, etc) and also specifically for a ADO (which bears a SBF). 

Solution: find a simple way of add a theme hint for SBF bearing Content Types (easy, code is there) but also, for specific JSON types. That way we can give people close to design and aesthetics the chance to shine. Beta3 task

PR #78, opened 2020-02-24
Issue-74: Add theme suggestions for Strawberryfield nodes

marlo-longley:
Addresses #74 

### What's new?

Screen Shot 2020-02-24 at 9 16 00 AM

Add hook in `strawberryfield.module` to suggest theme templates for strawberryfield bearing nodes. 
1. Always adds `page--node--strawberryfield` if SBF is present
2. Also conditionally adds  `page--node--strawberryfield--<type>` e.g. `page--node--strawberryfield--image` based on the type `as:` keys of the node.

### How to Test

Turn on Twig template debugging in `sites/default/services.yml` and set `twig.config` debug key to true. Then refresh a node page and see the hooks.

@DiegoPino is this what you were thinking?

### DIFF
--- a/strawberryfield.module
+++ b/strawberryfield.module
@@ -259,3 +259,47 @@ function strawberryfield_s3fs_url_settings_alter(array &$url_settings, $s3_file_
   //$account = Drupal::currentUser();
   //$url_settings['custom_GET_args']['x-user'] = $account->getAccountName();
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter() for node templates.
+ * @param array $suggestions
+ * @param array $variables
+ */
+function strawberryfield_theme_suggestions_page_alter(array &$suggestions, array $variables) {
+  $node = \Drupal::routeMatch()->getParameter('node');
+  $sbf = \Drupal::service('strawberryfield.utility')->bearsStrawberryfield($node);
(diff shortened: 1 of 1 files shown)


DiegoPino:
@marlo-longley something like that, but not that approach. There are a few issues with how you are tackling this issue

- Its too computational expensive. You parse a whole node, and then iterate over all JSON keys. That is quite CPU expensive and feels like an overkill to accomplish theme suggestion.

- The hook you are using is working on a level that is not really needed (you can argument in favor of this and i'm open to hear that). https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/function/hook_theme_suggestions_alter/8.2.x Seems like a better hook for this. Or even more exact https://api.drupal.org/api/drupal/core%21modules%21node%21node.module/function/node_theme_suggestions_node/8.2.x where you have direct access to a Node inside the `$variables` array.

- You are using the wrong JSON key to make the decision. This is really the most important point in my issue. `type` key is the one that drives display and also speaks about the ADO classification. the as:keys are about media types, attached files. The promise of Archipelago is exactly that media type does not define what the thing we are describing is, or not exclusively. Any ADO could eventually have any of those` as:mediatype` keys and that would not make a clear distinction of what an ADO is representing really. Or seeing in a different way, a metadata only ADO will become invisible to theme suggestion.

PR #85, opened 2020-05-08
ISSUE-84: Only notify user of Metadata driven title change if change is noticeable

DiegoPino:
# What is new?

Basically, don't tell the user everytime it saves a Node/ADO via a Webform that the title changed if its still the same. Mostly cosmetic, but was driving me insane. Also, if the change happened with an actual title field visible and not only driven by webform, don't notify the user because the Event Subscriber basically does nothing (sets Force Update to false and skips the logic, good)

@giancarlobi @mitchellkeaney this is just a heads up so you know i did this. Will merge later today


mitchellkeaney:
Truly thought there was a reason behind the notification of the title change and always wondered why it was this way. Always forgot to ask/bring it up, but i'm glad to know I'm not insane! 
thanks for the head's up about the change.

Issue #88, opened 2020-06-02
JSON:API Drush command wrapper


DiegoPino:
# What is this?

Just a way of helping people and myself to ingest/patch NODES faster via the JSONAPI. 
I will add a drush command that allows an arbitrary path/wildcard/filename to be passed as attachement, a JSON file to be passed as payload and credentials. Drush command will run all the needed JSONAPI calls, double encode, fetch responses and stuff to ingesting new objects via the API is simpler and less convoluted.

DiegoPino:
See also https://groups.google.com/forum/#!searchin/archipelago-commons/json$20api%7Csort:date/archipelago-commons/IN4gk-qwjuo/AdagXXm6BgAJ

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/b2517b4f9fcf274e3905b3503d98f6702efb0f77 and https://github.com/esmero/strawberryfield/commit/925fb58372149056ef3c0b76ab5420c08e6fa150

Issue #86, opened 2020-05-11
Extract basic File Metadata on File Persister Service and push as flavor


DiegoPino:
# What is this?

Basically a quick hack, don't do this at home kids for File Description extraction when dealing with ADOs/Archipelago Digital Objects with Small Binaries/not many Binaries and that don't have the luxury (yet) to have strawberry_runners doing all their cool ReactPHP Async background processing for you.

This adds a WIP Method to the filePersisterService that will process/move/to/tmp files, run exiftool and (UK) Pronom extraction (FIDO) and push back all the insane amount of data into the JSON.

# What is also needed?

- A check that says , hey! You are adding like 3 files, Ok, i can process in sync/realtime, but hey, those are 2000 pages of poetry and thoughts, no way i'm doing that, install/setup strawberry_runners for that, i will only do it for the first, 3? that are less than 100 Mbytes. OR so.

- Move Checksumming which runs on the main method into this ::getBaseFileMetadata() (oh.. you have not seen the pull yet.. well, linked down!)

- Add a Config Form for FIDO and EXIFTOOL. Future i want to use https://github.com/richardlehane/siegfried also

DiegoPino:
@giancarlobi when you have a moment i would like to discuss the following:
- I have been testing EXIF extraction on files and wondering now, how efficient/useful it is to actually keep all around in the SBF. Even when 10 images, the number of elements can grow too much. Its not the size, but the fact that we have multiple levels, and to be honest, exif is already! carried around in the actual file, so wondering what elements (and how can be define/decide) are the ones we want to extract. Or we want to simply dump them directly into a file (like the flavors we did last year) and let solr deal with them? If so we need to finish that code (urgent)
-  FIDO is ok, quite fast and won't take too much space.

The question i have now is how we can measure performance and impact. It most of the times not a server issue, but a UI issue when using the webforms/ajax, and still i don't know if that is even an issue or not. But i want to be able to provide people with options.

giancarlobi:
@DiegoPino it could be on wednesday, is it ok for you?

DiegoPino:
@giancarlobi i have a strange week ahead. Too many calls and teaching and meetings and development 😢 . Maybe we can start simply in the chat here and figure out a call closer to Friday? Also good afternoon for you!

giancarlobi:
@DiegoPino ok to start here and eventually a call on Friday. I'm little busy so I'll start posting here tomorrow. Have a nice day, amigo

[...]

PR #89, opened 2020-06-03
ISSUE-88: WIP of a JSON API Drush command wrapper

DiegoPino:
## This commit adds:
See #88 

- A wrapper drush command (example call)
Where the main argument is a pure SBF JSON (for now because it can/should also deal with a FULL ready to be posted JSON API data). It takes a --files argument with a folder of files or a single file (need to check if this is working, this is all quite precarious still), rest are the bundle, the user and pass for jsonapi and a moderation state if people want that to be set.

Right now this actually works and ingests a few files, then the node and connects the dots. But! Since `as:image` etc are still generated by the webform and not as an eventSubscriber, after the creation you still need to go the node, edit via we webform as such. Which is terrible. So next commit will fix that finally!

All this requires also
 - `docker exec -ti esmero-php bash -c "php -dmemory_limit=-1 /usr/bin/composer require drupal/jsonapi_earlyrendering_workaround; drush en jsonapi_earlyrendering_workaround;"`
For some reason (documented in many places like https://www.drupal.org/project/drupal/issues/3072076 there seems to be a render context missing and cache rendering stuff (long expl) is leaking giving us  bad error message at the  end, even when the ingest actually works!

So i added a check, this is a soft dependency since i don't want to depend on that contrib module, but drupal forces me into this situations.

Also @giancarlobi **i need help** with this script. Testing but also making it better. It feels basically just too precarious right now and i need to make it stronger. Will share demo assets for testing after my morning teaching sessions. That is also why i'm making this agains ISSUE-86 and not directly against Beta3. Once tested i can move forward.

This also adds some soft dependencies that should get solved once JSON-API fixes a weird issue with some leaking cache metadata (early rendering, means some code renders something before the response from JSON-API goes back but no caching happens for that). To be honest, why in the hell would you want to cache a POST request to create an entity??

## Steps to test this
- Make sure you have Drush 9 or 10. Not 8, 9 or 10.
- Unzip this folder in archipelago's cloned github repo inside the `d8content` folder
[ados.zip](https://github.com/esmero/strawberryfield/files/4725063/ados.zip)
That will create an `ado` folder with a json file inside an another folder which will contain a poster of trilobites.
- (if on Archipelago-deployment docker, if not you are already an expert and know what to do here1) 
```Shell 
docker exec -ti esmero-php bash -c "php -dmemory_limit=-1 /usr/bin/composer require drupal/jsonapi_earlyrendering_workaround; drush en jsonapi_earlyrendering_workaround; drush cr`
```

- Checkout this branch which also contains the same changes that ISSUE-86 has.
- drush cr again.
- Run the drush command like this (or adapt if not running via docker)
```Shell 
docker exec -ti esmero-php bash -c "drush archipelago:jsonapi-ingest /var/www/html/d8content/ados/ado-photograph.json --bundle=digital_object --uri=http://esmero-web --files=/var/www/html/d8content/ados/ado-photograph-bin --user=jsonapi --password=jsonapi --moderation_state=published"
```
You can also provide an --uuid=someuuid if you want to.  To avoid double ingesting the same one

That should
- upload the single image in that folder
- Take the full expanded JSON, compress, complemente, build an JSON API request, ask for an UUID and create a new Digital Object.

- You will see a LOT of output and that needs to be cleaned up.
- Script is not smart enough yet, so does not do all the checks needed! (HELP)
- Script will upload the same image over and over, but none of them will be attached until the node is processed
-  Check in your site for the content menu, a new published content will appear

@TODO: move the `as:` structure generator out of the webform handler into an event. That way persisting the file and extracting as is done on ingest via JSON-API too.


giancarlobi:
@DiegoPino A first question: the user set into drush command (i.e. jsonapi) have to exit into drupal users? can I use an admin user instead?

DiegoPino:
@giancarlobi since the Drush script calls the JSON API directly via CURL (so you can, e.g ingest even on a remote server) it needs to be any user and password that has HTTP/HTTPS access to the jsonapi endpoint. If your admin can then Ok. It does not use the drupal user (in terms of boostrapping and loading the user via PHP) because i need to pass the credentials via CURL. We could, eventually also in the future use cookies or JWT to make the call. But for now basic auth seems "perfect"!

giancarlobi:
@DiegoPino sorry for my next questions:
1) How/where I set a user able to access jsonapi endpoint?
2) in configuration Json:API (admin/config/services/jsonapi) Do I need to set "Accept all JSON:API create, read, update, and delete operations." or we bypass that setting?

DiegoPino:
@giancarlobi all good. yes to all what you say. I would create a specific user for the API that has, e.g,  only permissions to create digital objects and also to use the drop box file field! (its a special permission we added)  and that is why the example uses the jsonapi user. And yes, once you have that, you need to enable it. 

Idea: Since jsonapi should NOT be exposed to CRUD from the outside we could
- Firewall it
- Enabled via drush! and disable it when done! (automatically here!)
- Proxy it
Ideas?

giancarlobi:
> @giancarlobi all good. yes to all what you say. I would create a specific user for the API that has, e.g, only permissions to create digital objects and also to use the drop box file field! (its a special permission we added) and that is why the example uses the jsonapi user.

Thanks, I'll try that

> And yes, once you have that, you need to enable it.
> 
> Idea: Since jsonapi should NOT be exposed to CRUD from the outside we could
> 
>     * Firewall it
> 
>     * Enabled via drush! and disable it when done! (automatically here!)
> 
>     * Proxy it
>       Ideas?

I think 1 and 2 are good, together:
1) onto webserver drop connections to /jsonapi from unwanted IPs
2) enadle/disable by drush (i.e. we can include it in a main script)

giancarlobi:
@DiegoPino I checkout ISSUE-88 over beta 2, CR but ... running drush command I receive a : `There are no commands defined in the "archipelago" namespace.` Any ideas?

DiegoPino:
@giancarlobi you running this via docker or natively? in both cases you need a drush cr but if via docker the command needs to run via a docker exec because drush can not bootstrap from your storage only.

giancarlobi:
@DiegoPino I run natively and I executed a drush CR after checkout.

And also I have drush 9:
```
sudo -u www-data /var/www/archipelago/vendor/drush/drush/drush --version
Drush Commandline Tool 9.5.2
```

DiegoPino:
@giancarlobi so strange. Maybe 9 is the issue?
On docker i get this
```Shell
docker exec -ti esmero-php bash -c 'drush version'
 Drush version : 10.1.1 
```

So maybe i'm confused and it needs 10?

I am confused... sorry friend! It is stuck to 10! 😭 

https://github.com/esmero/strawberryfield/pull/89/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R27

giancarlobi:
@DiegoPino don't worry. I have to figure out how to update to 10 then I check this. Thanks !!!

giancarlobi:
@DiegoPino With drush 10 (I updated) no more error with drush command but running the drush command (as admin, I have to figure out all permissions that jsonapi user needs) **no** new content appears and I receive this error log:
```

Type | php
-- | --
Thursday, June 4, 2020 - 17:12
Admin
http://archipelago.byterfly.eu/jsonapi/node/digital_object
 
Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException: The attribute moderation_state does not exist on the node--digital_object resource type. in Drupal\jsonapi\Normalizer\ContentEntityDenormalizer->prepareInput() (line 70 of /var/www/archipelago/web/core/modules/jsonapi/src/Normalizer/ContentEntityDenormalizer.php).


```

giancarlobi:
@DiegoPino It works if I remove --moderation_state=published from command line. Great !!!

DiegoPino:
@giancarlobi ohh!! You don't have that field/module in your deployment. Ok, that makes sense. That is another check i need to add, that should not even be an option, maybe i just need a more generic approach for those.!

DiegoPino:
@giancarlobi i will check your site to see how publish (probably the 'status' field) is setup. Moderation is part of the worklow module and not every Content type/bundle will have it. YOu are right. Still a lot to fix and code. This is good. Will do more and let you know tomorrow in the morning!

Cheers and thx so much

giancarlobi:
@DiegoPino a minor issue, I think not related with this PR but I found it testing this, the default fido directory: on ubuntu installing fido as main instruccions, the executable is into /usr/local/bin (not /usr/bin as exiftool) so maybe we have to change here https://github.com/esmero/strawberryfield/blob/c00ae2bdd479b129e8d79febd8d9a3c32f378f76/src/StrawberryfieldFilePersisterService.php#L863 from 
$fido_exec_path = $this->config->get('fido_exec_path') ?: '**/usr/bin/fido**';
to 
$fido_exec_path = $this->config->get('fido_exec_path') ?: '**/usr/local/bin/fido**';

I know we'll can set that into configuration page when available.

DiegoPino:
@giancarlobi totally. I agree. I need to expose those 2 binary paths via a form/settings and add also the configurations to the schema. I was just reluctant to do so until really having all converted into an event call instead of how it happens now on webform submission.

Also: i feel i need to have some type of "tools" tab "managment" or anything like that (ideas for names) on each ADO where people can re-run this events.

DiegoPino:
@giancarlobi also, question to you:

What other features do you think are useful in this drush script?

Some ideas:
- Checking if a File already exist and if so just connect it. (not sure how useful!)
- Allow a full ADO deposited JSON to be ingested. I feel this one is very very useful!
 Your ideas?

Thanks!

giancarlobi:
> @giancarlobi i will check your site to see how publish (probably the 'status' field) is setup. Moderation is part of the worklow module and not every Content type/bundle will have it. YOu are right. Still a lot to fix and code. This is good. Will do more and let you know tomorrow in the morning!
> 
> Cheers and thx so much

I think I have workflow on my Archipelago, maybe with incorrect settings?
image

giancarlobi:
> @giancarlobi totally. I agree. I need to expose those 2 binary paths via a form/settings and add also the configurations to the schema. I was just reluctant to do so until really having all converted into an event call instead of how it happens now on webform submission.

Totally agree

> Also: i feel i need to have some type of "tools" tab "managment" or anything like that (ideas for names) on each ADO where people can re-run this events.

I think management is well explaining what people can do with that.

DiegoPino:
> > @giancarlobi i will check your site to see how publish (probably the 'status' field) is setup. Moderation is part of the worklow module and not every Content type/bundle will have it. YOu are right. Still a lot to fix and code. This is good. Will do more and let you know tomorrow in the morning!
> > Cheers and thx so much
> 
> I think I have workflow on my Archipelago, maybe with incorrect settings?
> image

Maybe? Maybe it is not being used by the Digital Object Content Type? It can also be the fact that content moderation has some big flaws (well.. everything has eventually some problems, not their fault really) so i installed this module:
 docker exec -ti esmero-php bash -c "php -dmemory_limit=-1 /usr/bin/composer require drupal/moderated_content_bulk_publish:^1.0"

So people can change the content moderation status batch using the admin/content view. (not possible).

One way you can check if its your config or this module.
When publishing draft. Does the new object appear also under the "Moderated Content" tab?
If not, means workflow is probably not working. I will check. Doing the config form now!

giancarlobi:
> @giancarlobi also, question to you:
> 
> What other features do you think are useful in this drush script?
> 
> Some ideas:
> 
>     * Checking if a File already exist and if so just connect it. (not sure how useful!)
> 
>     * Allow a full ADO deposited JSON to be ingested. I feel this one is very very useful!
>       Your ideas?
> 
> 
> Thanks!

@DiegoPino what about set "View mode" ? 

In addition, if I pass an UUID, the ingested object completely replace the existing? or merge it?

DiegoPino:
@giancarlobi good question. View mode should be an option people can setup/if they want! 
And also, UUID question. The current drush script (ingest) is for creating new ones, so if you pass the UUID and the UUID is there is should/and will complain.

So i want to have another, different drush alias for patching. But you can convince me, maybe we can use the same with an option, update if present, that has true/false, but is false by default,
Also: I wish i could give each file an UUID and depend on drupal for that, That way we can also update/patch existing files and replace them!

giancarlobi:
> > > @giancarlobi i will check your site to see how publish (probably the 'status' field) is setup. Moderation is part of the worklow module and not every Content type/bundle will have it. YOu are right. Still a lot to fix and code. This is good. Will do more and let you know tomorrow in the morning!
> > > Cheers and thx so much
> > 
> > 
> > I think I have workflow on my Archipelago, maybe with incorrect settings?
> > image
> 
> Maybe? Maybe it is not being used by the Digital Object Content Type? It can also be the fact that content moderation has some big flaws (well.. everything has eventually some problems, not their fault really) so i installed this module:
> docker exec -ti esmero-php bash -c "php -dmemory_limit=-1 /usr/bin/composer require drupal/moderated_content_bulk_publish:^1.0"
> 
> So people can change the content moderation status batch using the admin/content view. (not possible).
> 
> One way you can check if its your config or this module.
> When publishing draft. Does the new object appear also under the "Moderated Content" tab?
> If not, means workflow is probably not working. I will check. Doing the config form now!

@DiegoPino SOLVED !!!
I had to set here (I'm really sleeping!!):
image
So also drush command with --moderation_state=published is working good!!

giancarlobi:
> So i want to have another, different drush alias for patching. But you can convince me, maybe we can use the same with an option, update if present, that has true/false, but is false by default,
> Also: I wish i could give each file an UUID and depend on drupal for that, That way we can also update/patch existing files and replace them!

@DiegoPino I fully agree with you and I dont't want convince you because I think it's better two commands: 1 for new ingesting (error if UUID already exists) and 1 for update

DiegoPino:
@giancarlobi i will remove the default value for the binaries and allow people to set their own in the form (there i will add a default value  but just as a hint). I will also  add a toggle that allows that processing to be enabled/disabled. And disabled by default, but all this will be setup correctly, enabled, etc, on an docker archipelago-deployment. We will need for sure more documentation. Working on the forms now and some validation of the exec paths, moving services around. Thanks

giancarlobi:
> @giancarlobi i will remove the default value for the binaries and allow people to set their own in the form (there i will add a default value but just as a hint). 

Great!!

> I will also add a toggle that allows that processing to be enabled/disabled. And disabled by default, but all this will be setup correctly, enabled, etc, on an docker archipelago-deployment. 

Double Great!!

> We will need for sure more documentation. Working on the forms now and some validation of the exec paths, moving services around. Thanks

@DiegoPino Thanks you for this fantastic work!

DiegoPino:
@giancarlobi if you are still around, i need your advice.
I want to verify if the binaries,exifinfo and fido exits and can be run. I have the code.
But not sure how often to verify. In terms of preservation and making things robust i should check everytime i want to run them. There could be unexpected permission changes on the server, anything that is not controlled by the user and we should check and also, even with verification on the form settings can always be set via YAML. But then  i'm worried about performance. Should i check once per Object? Only on settings form save? Or on every file, means sometimes multiple times per object.

Code i will run is

```
 public function verifyCommand($execpath) :bool {
    $iswindows = strpos(PHP_OS, 'WIN') === 0;
    $execpath = trim(escapeshellcmd($execpath));
    $test = $iswindows ? 'where' : 'command -v';
    return is_executable(shell_exec("$test $execpath"));
  }
```

giancarlobi:
@DiegoPino I think we have to check A) on settings form save and B) eventually before runners, not on every file. The assumption is: if something change onto the server then I have to change settings. The B) is to avoid runners hangs and runners errors.
In addition each command has different flag for version:
- exiftool -ver
- fido -v
But probably you already know.

DiegoPino:
Ok. Makes sense. So for runners we will do that. For now i will move verification to form only. I was tempted to verify everytime because i HATE when islandora fails processing and all goes to hell invisibly. But true. Too much safety is not performant!!! Thanks!

DiegoPino:
@giancarlobi i added a lot of code here. I should really move this into more tickets but its basically just you and me and i don't think i have the energy to keep all separated. What i added is a Event Subscriber (perfectible of course) that does exactly the same (but in a more 'modern' way) than what was done via the webform handler: classify the as: structures and generate the metadata.
In case webform is used, for now, this new event subscriber will also run (so double processing, but second one is slimmer). My intention is to generate a signature on the JSON that tells this subscriber that it already was run to it has not to do again.

Sadly i can not simply use a temporary storage where i keep a list of all Subscribers that already executed and avoid  double because when the webform handler runs right now, there is still not even an UUID present (when creating an Object). but when the even subscriber run, the node UUID is already there. So matching which events belong to which content is not easy!

giancarlobi:
> @giancarlobi i added a lot of code here. I should really move this into more tickets but its basically just you and me and i don't think i have the energy to keep all separated.


@DiegoPino for me it's ok a single place, less energy, good.

> What i added is a Event Subscriber (perfectible of course) that does exactly the same (but in a more 'modern' way) than what was done via the webform handler: classify the as: structures and generate the metadata.

To check it, after merge PR ISSUE-88, I have to run drush ingest, right?

> In case webform is used, for now, this new event subscriber will also run (so double processing, but second one is slimmer). My intention is to generate a signature on the JSON that tells this subscriber that it already was run to it has not to do again.

Can we think in this Event Subscriber as a flavour? a special one, obviously but that run as other one? 
As other flavours when it end writes something into SBF-JSON to sign that it was run, am I right?

> Sadly i can not simply use a temporary storage where i keep a list of all Subscribers that already executed and avoid double because when the webform handler runs right now, there is still not even an UUID present (when creating an Object). but when the even subscriber run, the node UUID is already there. So matching which events belong to which content is not easy!

Can we use SBF-JSON to annotate it? (or maybe I completly misunderstood)

giancarlobi:
> @giancarlobi i added a lot of code here.

@DiegoPino sorry, I don't understand if you already added more code because my ISSUE-88 fork is aligned with this so probably new code not merged yet.

DiegoPino:
> @DiegoPino for me it's ok a single place, less energy, good.
Excelente!
> To check it, after merge PR ISSUE-88, I have to run drush ingest, right?
Yes or you can use the RAW metadata edit, e.g remove the as:images tag from the JSON string there and save. That should trigger the same.
> Can we think in this Event Subscriber as a flavour? a special one, obviously but that run as other one?
Yes. I would say yes. Since this one is CORE and is needed for every other one, i added it here  instead of in its own service. I actually would love to allow a queue/batch system for this one too in the future. So we can process hundreds of files in the same ADO.

> As other flavours when it end writes something into SBF-JSON to sign that it was run, am I right?
Yes. You are right. I think easiest way here is to actually sign into the SBF-JSON. 
> Can we use SBF-JSON to annotate it? (or maybe I completly misunderstood)
Yes. We can. Actually. I feel we should add all of them here, as a multi option
```JSON
   ],
    "as:generator": {
        "type": "Create",
        "actor": {
            "url": "https:\/\/play.archipelago.nyc\/form\/descriptive-metadata",
            "name": "descriptive_metadata",
            "type": "Service"
        },
        "endTime": "2020-06-11T15:18:39-04:00",
        "summary": "Generator",
        "@context": "https:\/\/www.w3.org\/ns\/activitystreams"
    },
```

src/Plugin/Field/FieldType/StrawberryFieldFileComputedItem.php
```diff
@@ -184,12 +188,12 @@ public static function generateSampleValue(FieldDefinitionInterface $field_defin
...
     // Prepare destination.
     $dirname = static::doGetUploadLocation($settings);
-    file_prepare_directory($dirname, FILE_CREATE_DIRECTORY);
+    \Drupal::service('file_system')->prepareDirectory($dirname, FileSystemInterface::CREATE_DIRECTORY);
 
     // Generate a file entity.
     $destination = $dirname . '/' . $random->name(10, TRUE) . '.txt';
     $data = $random->paragraphs(3);
-    $file = file_save_data($data, $destination, FILE_EXISTS_ERROR);
+    $file = file_save_data($data, $destination,  FileSystemInterface::EXISTS_ERROR);
```
DiegoPino, line 196: @giancarlobi this is to make everything D9 compatible

src/Plugin/Field/FieldType/StrawberryFieldItem.php
```diff
@@ -185,6 +185,61 @@ public function isEmpty() {
...
+    *   Array of data we want to save in the main property
+    *
+    * @return string|boolean
+    *   Returns either the correctly encoded string or boolean FALSE
+    *   Make sure you compare using === FALSE!.
+    *
+    * @throws  \InvalidArgumentException
+    *    If what is passed to ::setValue() is not an array.
+    */
+   public function setMainValueFromArray(array $jsonarray) {
```
DiegoPino, line 228: ::setMainValueFromArray allows me to save an array directly into e.g the field_descriptive_metadata field item, and it will convert to the string and also clear the caches `$this->flattenjson` and `$this->jsonjmesresults`

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -0,0 +1,208 @@
...
+    ];
+
+    $form['extractmetadata'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Should File level metadata extraction be processed?'),
+      '#description' => $this->t('If enabled, exiftool and FIDO will run on every file.'),
+      '#default_value' => !empty($config->get('extractmetadata')) ? $config->get('extractmetadata'): FALSE,
+      '#return_value' => TRUE,
+    ];
+    $form['exif_exec_path'] = [
```
DiegoPino, line 67: I need now an extra field here to allow arguments from exiftool, so people can have less EXIF tags extracted. I also want here the discussion point of #86 happening. Basically, don't put EXIF (and same for the flavour version) into the JSON itself, but into a file that we reference.

src/StrawberryfieldFilePersisterService.php
```diff
@@ -177,66 +214,72 @@ public function getDestinationUri(
     string $relativefolder
   ) {
 
-    // Default $relativefolder is a 3 char hash generated by a checksum algorith.
-    $current_uri = $file->getFileUri();
-    $uuid = $file->uuid();
+    if ($file && $file->isTemporary()) {
```
DiegoPino, line 217: THIS IS A BUG FIX @giancarlobi  (i need to make an issue so i can tag it)
If i remove the as:structure from an existing ADO's JSON and i save, the system will try to generate a new valid name for the file and will append an extra filetype. Basically: if the file is named
- giacarlo.jpeg
On save for the first time it will be named
-image-giancarlo-theuuid.jpeg
Next time, if i remove the as:structure
it will be named
-image-image-giancarlo-theuuid.jpeg but will still be left in the same place (instead of moved) because of this check
https://github.com/esmero/strawberryfield/pull/89/files#diff-376f27afff874dbd089c366d26b52febR568
Where we don't re-save files that are already persisted. Because, e.g another ADO could be using it, etc.
So with this fix, now, if the file was already persisted means someone (ourselves) already did our jobs and no need to renamed it. I can make this smarter of course!

src/Form/FilePersisterServiceSettingsForm.php
giancarlobi: @DiegoPino I fully agree with you to reduce exif output tags.
Exif output into file instead of SBF-JSON, can it make more complex to manage results? It's my only doubt.

src/StrawberryfieldFilePersisterService.php
```diff
@@ -907,7 +992,7 @@ public function getBaseFileMetadata(FileInterface $file) {
       $output_exif = '';
       $output_fido = '';
       $result_exif = exec(
-        $exif_exec_path . ' -json -q ' . escapeshellcmd($templocation),
+        $exif_exec_path . ' -json -q -a -gps:all -Common "-gps*" -xmp:all  -ImageWidth -ImageHeight -Canon -Nikon-AllDates -pdf:all -ee -MIMEType ' . escapeshellcmd($templocation),
```
DiegoPino, line 995: Added a few limits here in the meantime. This needs to go into the form.

DiegoPino:
@giancarlobi i also added  a few code comments, since the commit is so large. Just to explain what i'm doing here. Feel free to ask for more explanations or clarifications. Thanks

giancarlobi:
> @giancarlobi i also added a few code comments, since the commit is so large. Just to explain what i'm doing here. Feel free to ask for more explanations or clarifications. Thanks

@DiegoPino of course, better when I merged and checking how it works, as you know I'm an empiric man je je je

DiegoPino:
@giancarlobi also regarding where in the SB-JSON to keep track of all the processing/flavors/subscribers. We can have many event types, here is an example of a larger set of activities (same ontology i use to keep track of the webform)
https://www.w3.org/TR/activitystreams-core/#example-3

DiegoPino:
@giancarlobi sorry, last comment. My main concern with this new code is performance. For simple things, like a few images and a few PDFs all seems to perform the same way, but if you are adding like 100 images *via the UI* things can change. Not even sure how to start measuring this!

src/Form/FilePersisterServiceSettingsForm.php
DiegoPino: It can. It also means we need to revive the OCR work we started, since this will/should be indexed the same way. Putting it into a file is simple enough. Getting it back! is another thing. Probably the solution is in between, some in the JSON, others in the file. We need to discuss this for sure

giancarlobi:
> @giancarlobi sorry, last comment. My main concern with this new code is performance. For simple things, like a few images and a few PDFs all seems to perform the same way, but if you are adding like 100 images _via the UI_ things can change. Not even sure how to start measuring this!

@DiegoPino maybe by time consumed? that is, prepare 100 image of standard dimension then ingest and check how much time to complete. Probably dimensions affect time due to background storage and not service time execution. We can check with two set of 100 images with consistent different dimensions to isolate storage effect.

DiegoPino:
> > @giancarlobi sorry, last comment. My main concern with this new code is performance. For simple things, like a few images and a few PDFs all seems to perform the same way, but if you are adding like 100 images _via the UI_ things can change. Not even sure how to start measuring this!
> 
> @DiegoPino maybe by time consumed? that is, prepare 100 image of standard dimension then ingest and check how much time to complete. Probably dimensions affect time due to background storage and not service time execution. We can check with two set of 100 images with consistent different dimensions to isolate storage effect.

Yes. Wonder if we can programatically keep track of time? Idea: here
https://github.com/esmero/strawberryfield/blob/master/strawberryfield.module#L28-L33
We can start a timer, let all subscribers run, then count back time used.

giancarlobi:
> All in the same function. I can even ask each subscriber to log its own time?

@DiegoPino this means one log entry for each image? Does it make sense? I think this generates a lot of entry without sense and finally we read only the first and the last one ... or not?

DiegoPino:
No, a single log entry per Object. We can then simply calculate the size/number of images using the JSON metadata

src/Form/FilePersisterServiceSettingsForm.php
giancarlobi: > It can. It also means we need to revive the OCR work we started, since this will/should be indexed the same way. Putting it into a file is simple enough. Getting it back! is another thing. Probably the solution is in between, some in the JSON, others in the file. We need to discuss this for sure

@DiegoPino probably exif output is a little bit different from OCR output, I mean regarding output dimensions. For OCR I think JSON+file is ok, for EXIF we have to evaluate if putting result into file generate more CPU cycles to retrieve it than into only JSON

DiegoPino:
@giancarlobi i will be merging this into ISSUE-86 tomorrow to get things closer to release. I want to make a soft Beta3 release on Thursday afternoon to move to a few weeks of internal testing and then move finally to a 1.0.0 by end of July. I feel we only need  well working runners to get there! Sleep well!

src/Form/FilePersisterServiceSettingsForm.php
DiegoPino: I agree. maybe we simply not extract everything and can live with these. I was just getting too much overhead when dealing with EXIF heavy images and a lot of them (and also because of the flattening algorithm... but then now, thinking about this better, i actually can avoid flattening flv: elements!... should i open a ticket??)

strawberryfield.module
```diff
@@ -25,12 +25,38 @@ use Drupal\Core\Field\FieldItemListInterface;
...
+    $start_time = microtime(true);
+
     $event_type = StrawberryfieldEventType::PRESAVE;
     $event = new StrawberryfieldCrudEvent($event_type, $entity, $sbf_fields);
     /** @var \Symfony\Component\EventDispatcher\EventDispatcher $dispatcher */
     $dispatcher = \Drupal::service('event_dispatcher');
     $dispatcher->dispatch($event_type, $event);
+
+    if ($bench) {
+      $end_time = microtime(TRUE);
```
DiegoPino, line 46: I want not only to track this as a log, but inject into some setting/entity/even a queue? i can then process/avg?
pretty sure this processing is not even expensive. I was surprised really by how fast the filesystem was, specially S3! but probably the save event and the update events, because of MYSQL can be slower.

giancarlobi:
> @giancarlobi i will be merging this into ISSUE-86 tomorrow to get things closer to release. I want to make a soft Beta3 release on Thursday afternoon to move to a few weeks of internal testing and then move finally to a 1.0.0 by end of July. I feel we only need well working runners to get there! Sleep well!

@DiegoPino as you plan to merge this into ISSUE-86, does it make sense to check now or better I wait and finally check ISSUE-86 all together ?

DiegoPino:
@giancarlobi complex! But yes you are right, #86 is kinda a dependency for this one (defines the basics of the exif thing). So i should do that first. Then update the #86 open pull message to say it covers 86 and 88. But i also need to refine something in the drush command before that (which i plan on doing after teaching today).

PR #102, opened 2020-09-02
Add second parameter to converter

giancarlobi:


DiegoPino:
Great. For posterity: This is something we discussed last night with Giancarlo. He is installing archipelago on Drupal 9.x and now the Route Parameter Converter we built to upcast and UUID into a Node (`do/uuid`) requires two arguments in the constructor. This is a service so arguments are passed as injected dependencies.

DiegoPino:
@giancarlobi this has extra commits so i will just squash all into a single commit to make to keep the history clean. Thanks

Issue #65, opened 2020-01-30
show node Edit tab if user has access to any Form Mode


marlo-longley:
### What is wrong?
Right now, if a user role doesn't have permissions for the default form mode, but does have access to another form mode, the Edit tab is not displayed when viewing a node.

For example, in a recent project, a "Contributor" role has access to the "Contributor" form mode, but not default. They can only see the "Delete" tab, but not "Edit" above a node.

### How to fix
These tabs above the node are called local tasks in Drupal. See here:
https://www.drupal.org/docs/8/api/menu-api/providing-module-defined-local-tasks

Will create the following YML file: `strawberryfield.links.task.yml` and add code targeting nodes.
Another option is to use `strawberryfield_menu_local_tasks_alter`.

PR #111, opened 2020-11-04
ISSUE-65: Give users an edit button even if default form not enabled

DiegoPino:
See #66 and #65

This is @marlo-longley code that goes into 1.0.0-RC1. Needed to allow a decent UI


DiegoPino:
@alliomeria not asking for a Code review (we will get there) but mostly to any concerns/questions you may have

alliomeria:
@DiegoPino don't have any concerns/questions at this time, think this makes sense and will be helpful for Archipelago users

DiegoPino:
Thanks, will merge, we will have time now to test for real use cases

PR #119, opened 2020-12-14
ISSUE-114: OCR capabilities added, and the dream came true

DiegoPino:
# What are you adding here?

- A new field type via config (Giancarlo will ask me why. Because this allows us translations). Once installed via config sync, new custom fields and field types will appear. Your custom configuration option in Solr do not need any change
- A Highlight endpoint. It depends on certain Search API settings and using the correct OCR custom code in the config
PS: @giancarlobi I'm getting some strange Solr errors, probably we need to fine tune the  charfilters/tokenizers. It only happens with certain OCR/short words
- Also: probably my fault, but I'm getting different results when using 2Up v/s 1UP. 1UP is working great, but for some reason 2UP is failing on the display? Any ideas?
- An Search Alter to add our custom OCR highlight arguments


DiegoPino:
Looking good. Will merge!

Issue #121, opened 2020-12-09
Cleanup file structure for deleted files


alliomeria:
# What is this? :bug:

When files are deleted from an ADO, and a certain key becomes empty (such as "image"), the corresponding file structures ("as:image") are sticking around even after the file deletion is saved. 

This issue may have been introduced near Beta3; Related to `strawberryfield` module.

DiegoPino:
@alliomeria moved this to SBF since this modules handles the Persisting / Deleting files logic. Coding this now.

DiegoPino:
Fixed via https://github.com/esmero/strawberryfield/commit/8faca1362ae8adcf61e3e84dbd3e09ea7b3f7377

Issue #120, opened 2020-12-15
Bug: identify is not running on Draft ADOs


DiegoPino:
# Bug

.. and because of that, since the as:structure does get filled up, it never runs, even after publishing. This feels like a regression or a bad mad If() condition. Needs fix before 1.0.0-RC1 is made public
@dmer @alliomeria  will fix today

alliomeria:
@DiegoPino, I could not duplicate this issue in any Beta 3 instances I tested/checked.

DiegoPino:
@alliomeria thanks so much for testing. Clearly its a thing **I** 🤕 introduced in the most recent code. This helps a lot.

DiegoPino:
@alliomeria this may be the most stupid bug/not bug ever.

I tested and tested and could not replicated except on our 1.0.0-RC1 demo of today.
Look. Right is the stored/deployed config
Left. The config after pressing the config form for basic metadata extraction

So where is the issue?

WE "tried" to give the form "defaults" which does not mean that when you are seeing the form the values are actually saved!
The user has no way of knowing that the defaults there are not persisted and are only "illustrative".



![image](https://user-images.githubusercontent.com/6946023/102289844-a036c200-3f0d-11eb-93b1-871260652081.png)

DOUBLE MEGA FACEPALM

Of course previously processed files won't get (as the code is) any new tech metadata because we assumed them processed so you can save and save and nothing.

So. Solutions are:

- DO not put any defaults? Or put them as suggested (like in the background of the textfields but not filled up as values)
- IF we put defaults as they are Mark the ones that are not saved yet "as not stored"? Can be complex to explain

Plus:

actually ship with the settings saved... I did not....



![image](https://user-images.githubusercontent.com/6946023/102289734-5f3ead80-3f0d-11eb-93fe-0e27495ac062.png)


strawberryfield.filepersister_service_settings

patdunlavey:
Wouldn't the normal approach here be:
1. In the functions that depend on this value being set, where it has not been set, try to use a hard-coded default.
2. On the configuration form, pre-fill the default (same value as 1) if it has not been previously saved.

DiegoPino:
[...]

PR #122, opened 2020-12-17
ISSUE-121: Make file deletion / removal of as:structures better

DiegoPino:
See #121 
# What is new?

## On the StrawberryfieldFilePersisterService class
- Simple  public function `calculateAsKeyFromFile(FileInterface $file)` that uses the `mimetype` to get the `as:filetypes`, etc keys. Code was already there but since we needed to reuse it I made a method
-  Moved all mentions of 'as:' and 'urn:uuid' to CONSTANTS in the class. Cleaner, better.
-  New method that given a list of file ids removes any entry of those from the JSON inside the as:structure
public function removefromAsFileStructure(). I also, once removed decreases the Usage count since no other service has actually access to that info. And this is the right moment to do so. 
- Singe 1 count removal is applied using `$this->remove_file_usage($file, $nodeid, 'node', 1)` 
- Returns the original FULL JSON (not the previous one) with `as:filetype` entries removed.

PS: Since i was here added some Logging in case some of the basic identify, etc processors are empty and also double checked that in case an upload file has spaces we single quote them to avoid issues. This is related to #120

NOTE: @alliomeria @giancalro can you ingest a JP2 tomorrow and see if you get `flv:identif` data stored in your JSON? In one of my test machines identify could not and i hope its not because the shipped graphicsmagic lacks that decoder. If so I may want to compile Graphicsmagic again for the PHP 7.4 container

@patdunlavey. You will want to see this and ask/look since this service is quite heavy and important in the whole file parsing/cleaning/adding/keeping safe mechanic. And you will spend some time on this file in the future.

On the actual Event Subscriber (which uses the code explained before)

- We ask SBF for its original (previous saved data) and get the flattened values
- We ask for all the dr:fid values (means any file mentioned/referenced)
- During new file processing we keep a list of all File IDs added
- After that we compare old files with new files. Difference is Magically (hahaha) the ones that are not longer with us.

We call  `$this->strawberryfilepersister->removefromAsFileStructure(
                $to_be_removed_files, $fullvalues, $entity->id())` to remove those left over `as:filetype` from the JSON

Then instead of iterating and cleaning over the "newly processed" because guess what.. if you deleted all newly processed will have no iteration, we iterate over every possible key (only a few StrawberryfieldJsonHelper::AS_FILE_TYPE really) and check if there is previous data for that as:key, if there is new data for that same as:key. If new data is we use that as main and only complement with old left over data. Which one if we deleted that just now? Well, we can actually have URLS and EXTERNAL files not managed by us. But. That also needs to be mimic in the WEBFORM because that little rascal gets rid of those.

ALSO @alliomeria : only reason we have this processing on the webform is because of ESIE. We may totally want to disable that option and let this part deal with everything as originally planned. OR. We can make that processing simply a checkbox/option in the SBF harvester so the rest of the world does not need to suffer with extra ARRAY traversing.

How to test: 

- Create a new Object and add a few files. Save.
- Check if all is there and RAW metadata is OK and no warning in the logs

- Go back and remove all of a certain type of file (e.g documents): Save. 
- Check that both Files and as:entries are gone and also the corresponding e.g `as:document` is totally gone

- Go back and add more files: Save.
- Check if all is there and RAW metadata is OK and no warning in the logs

- Remove one of each type. Save
- Check if all is there and RAW metadata is OK and no warning in the logs

- Remove all Files. Save
- Check if NOTHING is left and rest of RAW metadata is OK and no warning in the logs

Now check in your admin/content/ for the deleted files and see if there are still Usages left

Usage is removed (-1) on each deletion
But also added +1 on each versioning (if all is good and I did not make any mistakes)

The idea is that NODE revision the contains a certain file adds usage
Every time a file is deleted we decrease USAGE and for extra security(nobody wants to loose a file) the Computed "drop_box" field we have keeps at ENTITY NODE level an extra connection which only gets removed if the WHOLE node and every revision are deleted.

### DIFF
--- a/src/Controller/StrawberryfieldFlavorDatasourceSearchController.php
+++ b/src/Controller/StrawberryfieldFlavorDatasourceSearchController.php
@@ -216,13 +216,17 @@ class StrawberryfieldFlavorDatasourceSearchController extends ControllerBase {
             $result_snippets_base = [];
             if (isset($field[$allfields_translated_to_solr['ocr_text']]['snippets'])) {
               foreach ($field[$allfields_translated_to_solr['ocr_text']]['snippets'] as $snippet) {
+                // IABR uses 0 to N-1. We may want to reprocess this for other endpoints.
+                $page_number = strpos($snippet['pages'][0]['id'], $page_prefix) === 0 ? (int) substr(
+                  $snippet['pages'][0]['id'],
+                  $page_prefix_len
+                ) : (int) $snippet['pages'][0]['id'];
+                // Just some check in case something goes wrong and page number is 0 or negative?
+                $page_number = ($page_number > 0) ? (int) ($page_number - 1) : 0;
                 $result_snippets_base = [
                   'par' => [
                     [
-                      'page' => strpos($snippet['pages'][0]['id'], $page_prefix) === 0 ? substr(
-                        $snippet['pages'][0]['id'],
-                        $page_prefix_len
-                      ) : $snippet['pages'][0]['id'],
+                      'page' => $page_number,
                       'boxes' => [],
                     ]
                   ]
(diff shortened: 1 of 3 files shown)


DiegoPino:
See #151 thinking of my code and mistakes here, I think my current approach is a bit overly complex.
Not sure I need to relay on previous stored values to identify which files are no longer there. A better approach would be to iterate over very as:filetype structure and check if dr:for still exists and contains its dr:fid, if not mark as to be removed or remove inmediatelly. Sometimes code needs to rest and needs to be read again by the same person that wrote it to see the light!

Issue #125, opened 2020-12-28
Move from Identify to pdfinfo for PDFs


DiegoPino:
# why?

because Identify on PDFs is a bottleneck.

a PDF processed (all pages) to extract Page dimensions takes almost a minute (128 pages)
pdfinfo takes less than a second. 

The difference is HUGE.

Also, we already ship with `pdfinfo`

What needs to be done?

We need to have the same output format.

## Identify
`identify -format 'format:%m|width:%w|height:%h|orientation:%[orientation]@' -quiet  sbr_somehash.pdf`
Which gives us for each line this
`@format:PDF|width:628|height:790|orientation:Undefined` which we split and JSON-ify

## pdfinfo

We need first to get the number of pages (a few ms)
`pdfinfo  sbr_somehash.pdf -quiet  | grep "^Pages:"`
We get this
`Pages:          124`
Now with that number (needs processing we can call)
`pdfinfo sbr_somehash.pdf -f 1 -l 124 |grep 'Page'`
which will give us a large list of two lines per page
Page  124 size: 630 x 794 pts
Page  124 rot:  0

This is also in the `ms` range of processing time

## what is needed?

I have to change \Drupal\strawberryfield\StrawberryfieldFilePersisterService::getBaseFileMetadata
to exempt PDFs and Ghostscript files from run through identify and instead run PDF Info.

@giancarlobi I wonder if that is the case I should keep the same output formate we use of identify?
Or would it better to go for its own flv:pdfinfo?

giancarlobi:
@DiegoPino Uhm ... probably more clean if we start with a new flv:pdfinfo. Also we can avoid to add `"format": "PDF",` and not sure if we need orientation too.
Can we imagine a jsonpatch to convert existing flv:identify ?

DiegoPino:
Good call!. Yes. we can do that.

DiegoPino:
I will also need to update all Twig templates we ship to ensure we are getting with from (both?) at the beginning and then deprecate identify for PDFs. Thanks!!

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/b70962c1e6fc1ac36b3e9b3927c4519a288584af

PR #126, opened 2020-12-28
ISSUE-125: Use pdfInfo for PDFs and PostScripts

DiegoPino:
# What is new?

ZUPER FAST PROCESSING!
(excited)

See #125 

## This changes how we process our PDF info:
1.- We now depend on pdfinfo (shipped with our docker container and configured by default so all good.
2.- Identify will only run only on Images and Movies
3.- `pdfinfo` data goes into `flv:pdfinfo` instead of `flv:identify`. To be discussed since it makes twig templating more "obscure" for users? Still. This way se save some space on unnecessary 'format'. Existing Objects can easily patched via AMI.
4.- We store both rotation(in degrees) and Orientation (to be exif complaint)

@giancarlobi can you test this?

**This is also WAY faster now than with identify!!** 🥳 

On an AMI batch ingest with 8 PDFs and other 8 other Objects.
We passed from 550 seconds to 18.61 sec!!!!! Almost 1 second per Object. And we are talking about 128+ pages per PDF.
That is 30!!! 30 TIMES FASTER! 30 TIMES Faster Kids. And this is from S3. So wow.

Happy!

### DIFF
--- a/src/StrawberryfieldFilePersisterService.php
+++ b/src/StrawberryfieldFilePersisterService.php
@@ -42,6 +42,7 @@ use Drupal\Core\Url;
 class StrawberryfieldFilePersisterService {
 
   const FILE_IRI_PREFIX = 'urn:uuid:';
+
   const AS_TYPE_PREFIX = 'as:';
 
   use StringTranslationTrait;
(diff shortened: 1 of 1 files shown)


DiegoPino:
PS: this also includes some DCS fixes so please excuse the mess. It was needed.

DiegoPino:
Super tested this. May as well merge @giancarlobi. If we find some issue we can fix on the branch.

giancarlobi:
@DiegoPino I think this works perfect: editing PDF remove old identify and add new pdfinfo. Also a little change to twig template Manifest 2.1 makes all working with IAB with new and old annotation. Also search in IAB is working good. Great job friend !!

PR #129, opened 2021-01-04
Add idle time to close loop

giancarlobi:
@DiegoPino Just a really draft idea to check if all queues are empty for at least 3 cycle. The 60 s timer check this and if all empty then close loop.

### DIFF
--- a/src/Commands/HydroponicsDrushCommands.php
+++ b/src/Commands/HydroponicsDrushCommands.php
@@ -81,19 +81,15 @@ class HydroponicsDrushCommands extends DrushCommands {
       });
     }
 
-//    $loop->addTimer(720.0, function ($timer) use ($loop, $timer_ping, &$done) {
-    $loop->addTimer(60.0, function ($timer) use ($loop, $timer_ping, &$done, &$idle) {
-
+    $loop->addPeriodicTimer(60.0, function ($timer) use ($loop, $timer_ping, &$done, &$idle) {
       // Finish all if all queues return 0 elements for at least 3 cycles
       // Check this every 60 s
-
       $all_idle = 1;
       foreach($idle as $queue_idle) {
         if ($queue_idle > 0) {
           $all_idle = 0;
         }
       }
-
       if ($all_idle === 1) {
         $loop->cancelTimer($timer_ping);
           foreach($done as $queue_timer) {
@@ -101,11 +97,19 @@ class HydroponicsDrushCommands extends DrushCommands {
           }
           \Drupal::state()->set('hydroponics.queurunner_last_pid', 0);
         }
-
       }
-
     );
 
+    $loop->addTimer(720.0, function ($timer) use ($loop, $timer_ping, &$done) {
+      // Finish all if 720 seconds are reached
+      error_log("All Done, 720 Seconds past, clearing the timers");
+      $loop->cancelTimer($timer_ping);
+        foreach($done as $queue_timer) {
+          $loop->cancelTimer($queue_timer);
+        }
+        \Drupal::state()->set('hydroponics.queurunner_last_pid', 0);
+      }
+    );
 
     $loop->run();
     Runtime::setCompleted();
--- a/src/Commands/HydroponicsDrushCommands.php
+++ b/src/Commands/HydroponicsDrushCommands.php
@@ -46,32 +46,69 @@ class HydroponicsDrushCommands extends DrushCommands {
     $active_queues = \Drupal::config('strawberryfield.hydroponics_settings')->get('queues');
     $done = [];
 
+    //track when queue are empty for n cycles
+    $idle = [];
 
     // Get which queues we should run:
 
     foreach($active_queues as $queue) {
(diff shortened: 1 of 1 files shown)


src/Commands/HydroponicsDrushCommands.php
```diff
@@ -46,32 +46,64 @@ public function hydroponics(
...
     }
 
-    $loop->addTimer(720.0, function ($timer) use ($loop, $timer_ping, &$done) {
-      // Finish all if 360 seconds are reached
-      error_log("All Done, 720 Seconds past, clearing the timers");
-      $loop->cancelTimer($timer_ping);
-        foreach($done as $queue_timer) {
-          $loop->cancelTimer($queue_timer);
+//    $loop->addTimer(720.0, function ($timer) use ($loop, $timer_ping, &$done) {
+    $loop->addTimer(60.0, function ($timer) use ($loop, $timer_ping, &$done, &$idle) {
```
DiegoPino, line 85: @giancarlobi, this time here runs only once, at 60 seconds. I think what we need to make this happen is that this time runs every 60 seconds (periodical). and we keep the security timer around.
If the periodical finds that all queues are idle, then it stops the security (720 seconds) one and itself. If not it keeps running.
The security one stops everything even the new periodical one that runs every 60 seconds.
We can make these times a setting too?

DiegoPino (commented):
I have a few comments. Let me know what you think @giancarlobi. Thanks for this!

src/Commands/HydroponicsDrushCommands.php
giancarlobi: @DiegoPino You are right, my error, I was thinking in periodic but it doesn't flow from brain to keyborad ....

giancarlobi: > We can make these times a setting too?

I don't know if we really need make it a settings, what about wait and eventually add it in the future?

src/Commands/HydroponicsDrushCommands.php
```diff
@@ -81,31 +81,35 @@ public function hydroponics(
...
         }
       }
-
       if ($all_idle === 1) {
         $loop->cancelTimer($timer_ping);
           foreach($done as $queue_timer) {
             $loop->cancelTimer($queue_timer);
           }
           \Drupal::state()->set('hydroponics.queurunner_last_pid', 0);
         }
```
DiegoPino, line 99: Great! A tiny little thing missing here. 
- After setting the state it needs to cancel itself. `$loop->cancelTimer($timer)`
- And it needs to cancel security one too, so the line `$loop->addTimer(720.0..` needs to be defined before the 60 second one and passed to the 60 seconds one so, something like `$securitytimer = $loop->addTimer(720.0..`. But... here is the complexity? How do we make this last timer suspend also the 60 seconds one?

giancarlobi: We have to use $loop->stop(); as here https://github.com/esmero/strawberry_runners/blob/25ee9fefbfc6d43942d6640fc930e2ef904f9676/src/Scripts/mainLoop.php#L348
As 720 probably will disappear in the future so I'd add $loop->stop(); to 60s timer and, at the moment, make 60s cancel 720s too. What do you think?

src/Commands/HydroponicsDrushCommands.php
```diff
@@ -92,22 +92,25 @@ public function hydroponics(
...
         $loop->cancelTimer($timer_ping);
-          foreach($done as $queue_timer) {
-            $loop->cancelTimer($queue_timer);
-          }
-          \Drupal::state()->set('hydroponics.queurunner_last_pid', 0);
+        foreach($done as $queue_timer) {
+          $loop->cancelTimer($queue_timer);
+        }
+        \Drupal::state()->set('hydroponics.queurunner_last_pid', 0);
+        $loop->stop();
```
DiegoPino, line 99: Hi, oh sorry my fault. I meant instead of $loop->stop() $loop->cancelTimer($timer). Just its own cleanup...

DiegoPino:
@giancarlobi good. I will test this thoroughly. I need to also push my changes to the settings forms. This may take me a few more hours but we will get there. The open question now will be the logging and even if monolog requires a site wide change, we may want to explore it since it gives us faster/less DB dependent logging which means also better performance.

giancarlobi:
> @giancarlobi good. I will test this thoroughly. I need to also push my changes to the settings forms. This may take me a few more hours but we will get there. 

No rush, friend.

> The open question now will be the logging and even if monolog requires a site wide change, we may want to explore it since it gives us faster/less DB dependent logging which means also better performance.

Today or tomorrow I can try to install monolog and set on my archipelago, giving you a full list of changes, so we can evaluate if is the right way.

DiegoPino:
@giancarlobi last thing! We may also want to clean up all the error_logs we have in SBRs..
Does not matter if we get rid them here only since each queue worker will still want to/hopefully/emit their own.
The logging thing is becoming more and more important so happy you can test Monolog in your server!

DiegoPino:
Ok. This is looking good. I will change the error logs to a drupal::logger in a next pull. Thanks @giancarlobi , great contribution!

Issue #134, opened 2021-01-10
Edge case, strange situation, and ADO without a SBF field / a fix.


DiegoPino:
# What?

Ok, this is super strange and requires more research, how it happened, why it happened.
@alliomeria during RC1 testing and content preparation managed to create an ADO without any SBF data. Not even the field itself. It is practically impossible since the SBF (field_descriptive_metadata) is required. The consequence of this is that during a second save, not with correct metadata, we fail on accessing the `->original->field_descriptive_metadata` during the AS Structure Event subscriber because we assume its going to be there. Well its not.

How to reproduce? No idea (yet) but the fix is simply to check if the previous revision in fact has a SBF before trying to act on it.

This requires more log checking and understanding how one can actually get around all the Data consistency in Drupal to not ingest a given field that is required, but, the fix I prepared is consistent and allows us to keep processing even in such strange event. 

More later on this, if we ever know what happened.

DiegoPino:
Fixed via https://github.com/esmero/strawberryfield/commit/fb5c0b2e7c660bc6372d5baa9a70f3f7fd2e92f2

PR #135, opened 2021-01-10
ISSUE-134: check if SBF exists on previous revision before fetching flat version

DiegoPino:
See #134 

This works around this un-explicable edge case.


DiegoPino:
Merging since this transparent for 99.999% of the users (at least I wish its so)

PR #140, opened 2021-01-20
Push 1.0.0-RC1 into main

DiegoPino:
#Just house keeping.

Eventually we really just want to get rid of all main/master non versioned branches. No need to have them

Pretty sure Master is messy here. May need to fix conflicts


DiegoPino:
I had to reset and force push master to allow RC1 to merge clean into main. Was quite messy. Also @alliomeria @pcambra @giancarlobi time to move over old `master` branches. No masters, no gods, no bosses, no idols and no heroes. My 2021 motto! For now will rename to `main` once I get all synced, but we want to really stop using them. Just confusion. Once we are over 1.0.0 we can start getting rid of them since NOBODY uses them

PR #142, opened 2021-01-28
ISSUE-141:Added a generic method in the utility service that allows for counting search results and can be reused in the rest of the codebase.

### LINKED ISSUE #141: Add a generic service that allows for getting a count of Search API results
This issue complements https://github.com/esmero/format_strawberryfield/issues/118

pcambra:
Additions on this PR:

- New method `getCountByProcessorInSolr` in `StrawberryfieldUtilityService` that is used in format_strawberryfield: https://github.com/esmero/format_strawberryfield/issues/118
- Minor code standards changes on the `StrawberryfieldUtilityService` class.
- NEW interface `StrawberryfieldUtilityServiceInterface` to have a contract that we can break later (so we can tear apart the service in smaller ones).
- Changes on `StrawberryfieldFlavorDatasource`, some minor coding standards, some changes from public to protected on the DI getters/setters
- NEW interface `StrawberryfieldFlavorDatasourceInterface`


src/StrawberryfieldUtilityService.php
```diff
@@ -332,7 +284,50 @@ public function verifyDrush(string $execpath, string $home) :bool {
     return $canexecute;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCountByProcessorInSolr(EntityInterface $entity, string $processor, array $indexes = [], string $checksum = ''): int {
```
DiegoPino, line 290: Thinking loud here and being a little too strict, but for consistency we tend to make not required options by default NULL. If you are checking for $checksum as a string and because of how many things can be validate to nothing, I would either check for
strict $checksum !== '' or go this simpler route, where NULL can actually be omitted as an option but will cleanly (and also semantically be congruent) with the fact we are not giving any. Not a blocker. But reads in general better.

src/StrawberryfieldUtilityService.php
```diff
@@ -332,7 +284,50 @@ public function verifyDrush(string $execpath, string $home) :bool {
...
+    $count = 0;
+
+    // If no index specified, query all of them.
+    if (empty($indexes)) {
+      $indexes = \Drupal::entityTypeManager()
+        ->getStorage('search_api_index')
+        ->loadMultiple();
```
DiegoPino, line 297: Since this function is very specific to our Data Source and the fields we have there, it would make more sense to only use indexes where the data source (Strawberry Flavour Data Source) is setup to be indexed. There is some example code in SB Runners for that.

DiegoPino: Note: not a blocker. Since there is a filter of course for the data source in the query.

src/StrawberryfieldUtilityServiceInterface.php
```diff
@@ -0,0 +1,148 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\search_api\Entity\Index;
+
+/**
+ * Defines an interface the common strawberry field utility service
+ *
+ * @ingroup strawberryfield
+ */
+interface StrawberryfieldUtilityServiceInterface {
```
DiegoPino, line 14: This makes me happy. Super needed

DiegoPino (changes requested):
Excellent work @pcambra. My only request would be to make default arguments NULL (specially checksum not an empty string but NULL) because I feel its cleaner when checking for its presence or not but also because we already have that in the project as almost an unspoken standard. Makes internal checks more consistent and carries (yes performance but at nano/pico level) less data around.  I appreciate you cleaned up some container level requests/code/use statements and specially the addition (🥳) of an Interface!!

src/StrawberryfieldUtilityService.php
```diff
@@ -159,38 +159,22 @@ public function bundleHasStrawberryfield($bundle = 'digital_object') {
...
-   * @return \Drupal\Core\Entity\EntityInterface[]
-   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
-   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * {@inheritdoc}
    */
   public function getStrawberryfieldConfigFromStorage($bundle = 'digital_object') {
-
-    $field = $this->entityTypeManager->getStorage('field_config');
-    $field_ids = $this->entityTypeManager->getStorage('field_config')->getQuery()
+    $field_config = $this->entityTypeManager->getStorage('field_config');
```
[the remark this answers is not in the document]
pcambra: That makes a lot of sense, but I use Checkhov variables, only put them there if I'm going to use them :D

src/StrawberryfieldUtilityService.php
pcambra: I don't have a strong preference, I tend to put the "empty" value for the data type (array is [], string ''...) but you're right that if we have a standard, I should follow it.

src/StrawberryfieldUtilityService.php
pcambra: Makes sense, I've taken the chance to sort out a little bit the datasource class

pcambra:
Note: force push is to rewrite the commit message and link it with this issue

DiegoPino:
@pcambra force pushing in your own forks is totally fine, and I normally do not care too much about the commit messages. The ISSUE title part is for the pull itself, which you can always manually modify even her via the web. Depending on how many simultaneous open pulls we have against the same branch and if there is overlap on the same files, I may decide on doing a squash and merge or a normal one. In the case of the first, the individual commits will just become messages.

src/StrawberryfieldUtilityService.php
DiegoPino: Thanks a lot!

src/StrawberryfieldUtilityService.php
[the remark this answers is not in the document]
DiegoPino: using the name `Checkhov` will always make me happy!

DiegoPino:
@pcambra the refactor on the Data Source is larger (because of line shifts, etc) and I need to test it. Could you please add a note in your Pull message that says that this is also done? Just for clarity we do not repeat of course the whole ISSUE info (ISSUE is more about explaining what is needed + some hints on how) and the pull message is more about how it was done. Even if short.

All related to the services is perfect now! Once I get my hands on that data source change I will approve (if no errors) and merge. Good work!

DiegoPino (approved):

DiegoPino:
@pcambra this looks good. I have another tiny thing to discuss with you here and a comment which is not really a thing.
Discussion: there is an issue where the Search API tries to get a View Mode and render it for this data source even if we return empty View Modes.
Comment: I know drupal proposes we use `null` instead of `NULL`. I may want to skip that rule and use `NULL` all over the place. Makes reading easier and its way more Symfony. Thoughts?

pcambra:
> I may want to skip that rule and use NULL all over the place. Makes reading easier and its way more Symfony. Thoughts?

My thoughts are "null" :) if we want to use bool constants in uppercase that's fine with me

>  Discussion: there is an issue where the Search API tries to get a View Mode and render it for this data source even if we return empty View Modes.

We discussed this today, I'll be happy to review this in depth, maybe in a spin off from this?

DiegoPino:
@pcambra 🥇 first merge in and immortalized in code!  Good work.

Issue #143, opened 2021-02-05
Extend EntityJmesPathNameProvider to cover all supported Entities by Solr


DiegoPino:
# The story

We coded this cool thing a while. Allows IDs in the JSON to be casted into proper dynamic entity references and because of that Solr can then index all, including hierarchies, etc.

Right now its "hardcoded" for `node` entities, because we only really care for ADO to ADO relationships, but can be extended by adding an extra setting and some wrapper code to anything.

# The fix (super simple)

## Setting would go here:

https://github.com/esmero/strawberryfield/blob/1.0.0-RC2/src/Plugin/StrawberryfieldKeyNameProvider/EntityJmesPathNameProvider.php

And wrapper code here:

https://github.com/esmero/strawberryfield/blob/1.0.0-RC2/src/Plugin/Field/FieldType/StrawberryFieldItem.php#L175

Where instead of passing "node" we would pass the new setting, e.g taxonomy_term

As simple as that.

@patdunlavey @pcambra @alliomeria this would allow us to index into Solr more complex things. Not urgent right now but can be useful for people relaying on Taxonomies 

The list of things we can support is based on what Solr can index. So needs to come from this constant:

https://github.com/esmero/strawberryfield/blob/1.0.0-RC2/src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php#L33-L42, that said I may take that constant to some other place. Really not a good place live in.

patdunlavey:
@DiegoPino are you suggesting this instead of the idea I  proposed on Friday of handling entity id references during presave by loading the entity and writing the values of its fields into the object JSON (e.g. taxonomy term name, description, custom fields)? If I understand what you are proposing, it would make storing these values in the ADO JSON potentially superfluous, and have a the advantage of being dynamic (e.g. if the price for a price bundle taxonomy term changes).

patdunlavey:
@DiegoPino would you mind taking a look at the attached patch and tell me if I'm on the right track?
[issue-143.patch.txt](https://github.com/esmero/strawberryfield/files/5946900/issue-143.patch.txt)

patdunlavey:
Here's an easier way to review: https://github.com/Born-Digital-US/strawberryfield/compare/1.0.0-RC2...Born-Digital-US:issue-143?expand=1

DiegoPino:
Hi @patdunlavey its looking good. I just pushed a branch for you that has the schema change and an update hook.
[...]

PR #144, opened 2021-02-09
ISSUE-143 --- added entity type config to EntityJmesPathName, and cor…

patdunlavey:
This PR 

- adds an entity type selection element to the "Entity Reference JmesPath Strawberry Field Key Name Provider" form based on entity types supported by https://github.com/esmero/strawberryfield/blob/1.0.0-RC2/src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php#L33-L42
- replaces 'node' with the the selected entity type in \Plugin\Field\FieldType\StrawberryFieldItem::propertyDefinitions

I understand clearly what the form modification does. I'm less sure what the change in StrawberryFieldItem::propertyDefinitions actually does.

See https://github.com/esmero/strawberryfield/issues/143

### DIFF
--- a/src/Plugin/Field/FieldType/StrawberryFieldItem.php
+++ b/src/Plugin/Field/FieldType/StrawberryFieldItem.php
@@ -109,6 +109,7 @@ use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
        ->setReadOnly(TRUE);
 
      $keynamelist = [];
+     $plugin_config_entity_configs = [];
      $item_types = [];
 
      // Fixes Search paging. Properties get lost because something in D8 fails
@@ -139,6 +140,11 @@ use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
            }
            //@TODO HOW MANY KEYS? we should be able to set this per instance.
            $keynamelist[$processor_class] = array_merge($plugin_instance->provideKeyNames($entity_id), $keynamelist[$processor_class]);
+
+           // Store the configuration options for use later.
+           if(!empty($configuration_options['exposed_key'])) {
+             $plugin_config_entity_configs[$processor_class][$configuration_options['exposed_key']] = $configuration_options;
+           }
          }
        }
      }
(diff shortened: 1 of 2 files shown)


patdunlavey:
I'm not sure if this actually works to make referenced entity attributes and fields indexable in solr - it's likely that don't understand how to configure all the moving parts. We should probably screenshare to make sure I'm performing all the correct steps.

I tried it out with two taxonomy select fields, one of which uses a vocabulary that has a custom field. I added the Entity Reference JmesPath Strawberry Field Key Name Provider (the webform field machine names seen in the ADO json together with the entity type). 

I'm able to add these fields to the solr fields, e.g. `field_descriptive_metadata:obj_adm_price_bundle:body:processed`. 
However:
- Custom fields do not appear in solr's add-field options. 
- When the object is indexed in solr, I see nothing for these fields - not the raw value nor the title or other fields. I only see the webform field name listed in `sm_str_flatten_keys`.

DiegoPino:
Hi, yes, let's do some live testing. I still need to review your code live on a machine, maybe there is something missing? Give me a few hours (sorry) but should be done by 16:00.

DiegoPino:
@patdunlavey in the meantime can you give the pull body (message) a bit of extra info about what is happening in the code? That will help future contributors a lot. Thanks!

patdunlavey:
@DiegoPino I tried to flesh it out just a bit, without going beyond what I actually understand!

patdunlavey:
It turns out my previous testing failures were a combination of various kinds of boneheadedness. I succeeded in getting this all working this morning, with the code in this Born-Digital:ISSUE-143 branch.
I ran code inspection in PhpStorm and it flagged no errors, though a number of warnings (not in new code). Do we care?

DiegoPino:
I do not care about every warning and normally won't expect anyone (except if very deep into the code) to touch more than it is required. Most important as we go further with more contributors are DCS. If you are OK I will merge this into ISSUE-143 and then merge into RC2. If you have any other concerns we can discuss them during our call today too. Thanks a lot, this is good work and will be useful for many other users.

DiegoPino (approved):
Good. All this is now working and no more code conflicts (because of me too)

patdunlavey:
I found one possible problem. When the webform's term select list is multi-value, the field does not get indexed in solr. The jmespath expression returns an array of values rather than a string.

DiegoPino:
@patdunlavey looking at it. Entity reference fields should allow arrays of entities. I may have missed a check? Does it also happen when nodes? (is member of?)

DiegoPino:
@patdunlavey some updates on this. This is a more complex refactor and I will open another issue for this one. let's roll with this current solution (even if its a single value indexed in Solr) until I find the solution. 

I will bring @pcambra into the issue too since I feel I will have to do some real work on understanding a better way of doing this.

DiegoPino:
See #146

Issue #151, opened 2021-02-20
Image persists from cloned object until saved a second time


aksm:
Not sure if this is expected behavior, but when I clone an object, remove the image from the original object, attach a new image, and save as the new object, both images persist. If I save the object a second time, the original image disappears and only the new image remains.

DiegoPino:
@aksm now that this is here: this is the issue

https://github.com/esmero/strawberryfield/blob/026847c1210e18d60e5c76c21f75f62a4d8a2633/src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php#L149-L153

I assumed (wrongly as I notice now) that new entities have  no `as:image`, `as:document` , etc JSON structures yet so do not need those to be cleaned up because file adding and removing at that stage (new) only generates changes in the actual JSON Keys that are fed by the "upload" elements.

But! When you clone an object, the already existing ones are brought over but its also marked as "new". So I skip cleaning that up

All this was made to avoid over processing but clearly I can not skip this test

Solution:

A bit more processing will be needed
Pseudo Code will be:

- For new entities only do expensive processing if `as:image` etc is there (that will make new Objects faster to process)
- If as:structures are there
  -  For each as:structure, check their `dr:for`key. If dr:for exists then check if the key still contains the File ID. If not, remove.

Since File persistence is quite important in Digital Repositories and I do not want (by mistake) loose a file, I will have to do some test+extra test on this. Still a few hours left before fixed!

Thanks @aksm, marking this as a Bug and will let you know when the pull is in.

DiegoPino:
@aksm a pull is in, if you want to test.
https://github.com/esmero/strawberryfield/pull/152

Simplest way may be to go to your `archipelago-deployment` folder:
```SHELL
cd web/modules/contrib/strawberryfield
sudo git fetch
sudo git checkout ISSUE-151
```

Test and let me know if saving is working. This is a larger refactor but it feels I got it this time right.

aksm:
That did the trick! Amazing work as always. Thanks!

DiegoPino:
Thanks to you @aksm!! We will merge this into RC2 tomorrow once testing is done

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/a489f68fd2c46c69bb6aaaeeb9d00bced8c82fd6

PR #152, opened 2021-02-22
ISSUE-151: Another different and better way of clearing/managing of `as:structures`

DiegoPino:
#151

# What is new?

Instead of reaching out to the original entity SBF field values(if any) we use the flat version of the current SBF to fetch all `dr:fid` (referenced File IDs). We also fetch `dr:for` (source Keys for those field ids) to make sure we can, in case of "emergency" rebuild our file mapping from there and we do if that is the case (new)

We compare effectively File IDs inside `as:filetype` with File IDs provided in user contributed Keys found in the `ap:mapping` structure (in that order). If there are differences, those differences are File IDs that do not longer exist/were not provided by the user. We clean those up from the `as:filetype` and also check  (new) if there are Files that do not longer exists in Drupal (edge case but can happen if you copied a full JSON between repos) and clean them up Brut Force (basically old fashion foreach over an foreach checking who is referencing not existing Files).

Code is cleaner and should make sense but i need testing.

How to test:

- Ingest a new ADO without files. 
- Clone the previous files ADO and save without adding new files.  
- Edit and add files. Save.
- Edit and remove files add new Ones. Save
- Edit and Remove all files. Save

- Ingest a new ADO with Files.
- Clone and Change files during edit. Save
- Clone again and remove all files and Save.

- AMI ingest a few objects via CSV.

- Delete Objects and check if Usage Count decreases or not

For all of these, make sure no errors or warnings happen and in each step all files are inside the RAW JSON and the proper JSON as:filetype structures exists, no extra files, no missing files.

This is Core to archipelago an any error/not caught exception may be our/your files doom! So please test thoroughly!

@aksm @alliomeria @pcambra please give this a look and feel free to ask what the hell is happening.

### DIFF
--- a/src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php
+++ b/src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php
@@ -10,6 +10,7 @@ use Drupal\strawberryfield\StrawberryfieldFilePersisterService;
 use Drupal\Core\Logger\LoggerChannelFactoryInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
+use Drupal\Component\Uuid\Uuid;
 
 
 /**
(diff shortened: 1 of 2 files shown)


alliomeria:
Hi @diegoPino--ran through the test series you outlined above. All went well, as:filetype structures corresponded with the right files, reflected removed/no files; no extra or missing files at any step. Usage count decreased as expected after AMI-ingested ADOs were deleted. Will test more tomorrow (see what happens with revisions, different AMI set with more files-per-object) and let you know if anything amiss occurs.

DiegoPino:
Gracias @alliomeria !

src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php
```diff
@@ -277,10 +259,20 @@ public function onEntityPresave(StrawberryfieldCrudEvent $event) {
...
+   *
    * @return array
    *   The cleaned/up $entityMapping
    */
-  private function cleanUpEntityMappingStructure(array $fullvalues) {
+  private function cleanUpEntityMappingStructure(array $fullvalues, array $for_from_as) {
+
+    // If not present of empty and we do have dr:for try to rebuild from that
+    // Its an edge cases but adds an extra level of dexterity/safety.
+    if (!isset($fullvalues['ap:entitymapping']['entity:file']) && !empty($for_from_as)) {
```
pcambra, line 273: If `$for_from_as` can be empty, I think it should be initialized to [] in the signature.

src/StrawberryfieldFilePersisterService.php
```diff
@@ -824,7 +822,46 @@ public function removefromAsFileStructure(
...
+
+  /**
+   * Removes a list of File IDs from the as:structure in a non optimal way.
+   *
+   * @param array $file_id_list
+   * @param array $originaljson
+   *
+   * @return array
+   */
+  public function removefromAsFileStructureBrutForce(
```
pcambra, line 852: I *think* if an method param is nullable, it should be at the end of the list

pcambra (commented):
A couple of small comments :)

src/EventSubscriber/StrawberryfieldEventPresaveSubscriberAsFileStructureGenerator.php
DiegoPino: Thanks. The caller does the checks here but for sanity I gave it a default

src/StrawberryfieldFilePersisterService.php
DiegoPino: In this case its really not nullable, since it needs to be an array, even if empty. For this function and its more precise counterpart I removed the defaults at all. Mostly because I did not wanted to change the signature's order. Thanks. Hope that addresses the problem

pcambra: It does, thanks!

Issue #153, opened 2021-02-25
Implement a different mimetype guess() approach


DiegoPino:
# What is this?

I may need some help with PROS and CONS here but my tests so far are resulting in success

Right now the ::guess function does this:

```PHP
/**
   * {@inheritdoc}
   */
  public function guess($path) {
    if ($this->mapping === NULL) {
      $mapping = $this->defaultMapping;
      // Allow modules to alter the default mapping.
      $this->moduleHandler->alter('file_mimetype_mapping', $mapping);
      $this->mapping = $mapping;
    }

    $extension = '';
    $file_parts = explode('.', \Drupal::service('file_system')->basename($path));

    // Remove the first part: a full filename should not match an extension.
    array_shift($file_parts);

    // Iterate over the file parts, trying to find a match.
    // For my.awesome.image.jpeg, we try:
    //   - jpeg
    //   - image.jpeg, and
    //   - awesome.image.jpeg
    while ($additional_part = array_pop($file_parts)) {
      $extension = strtolower($additional_part . ($extension ? '.' . $extension : ''));
      if (isset($this->mapping['extensions'][$extension])) {
        return $this->mapping['mimetypes'][$this->mapping['extensions'][$extension]];
      }
    }

    return 'application/octet-stream';
  }
````

You see that in this case the attempts are made from more general to more precise.
This leaves our warc.gz, and tar.gz out of the selection defaulting to application/gzip even if have provide our own mappings
In most cases extensions are mapped to a single mime but in Digital repositories people are quite aware of that and also we want to use precise mime detection to be able to decide if an asset qualifies or not be used by a certain formatter

My implementation (using our own existing service that provides also a reverse mimetype to extension) does this

```PHP
/**
   * {@inheritdoc}
   */
  public function guess($path) {
    if ($this->mapping === NULL) {
      $mapping = $this->defaultMapping;
      // Allow modules to alter the default mapping.
      $this->moduleHandler->alter('file_mimetype_mapping', $mapping);
      $this->mapping = $mapping;
    }

    $extension = '';
    $file_parts = explode('.', \Drupal::service('file_system')->basename($path));

    // Remove the first part: a full filename should not match an extension.
    array_shift($file_parts);

    // Iterate over the file parts, trying to find all matches.
    // For my.awesome.image.jpeg, we try and accumulate:
    //   - jpeg
    //   - image.jpeg, and
    //   - awesome.image.jpeg
    $qualified = ['application/octet-stream'];
[...]

PR #154, opened 2021-02-25
ISSUE-153: Deal with multi dot extensions (warc.gz, tar.gz) 

DiegoPino:
#What is this?

A different approach on getting the mimetype of an extension by accumulating all matches and returning the last one. If it is of course  slower because but we call this only during ingest.

See #153

@pcambra @patdunlavey @giancarlobi please look at this and fight me back. If not I will merge because I need precise warc.gz files!

### DIFF
--- a/src/StrawberryfieldFilePersisterService.php
+++ b/src/StrawberryfieldFilePersisterService.php
@@ -283,10 +283,25 @@ class StrawberryfieldFilePersisterService {
         $current_uri,
         PATHINFO_FILENAME
       );
+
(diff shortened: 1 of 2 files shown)


DiegoPino:
@ikreymer This solves also https://github.com/esmero/format_strawberryfield/issues/126

Tested and over tested. New Mimetype/Extension resolution + multi dot (two levels). This combines with another pull I have open for the replay.web embedded player in format_strawberryfield and one in https://github.com/esmero/strawberry_runners/pull/24

Will have all merged by tomorrow. Thanks!

PR #157, opened 2021-03-03
ISSUE-156: Custom Edit Route when "use default" is disabled fails to check if form mode is enabled

### LINKED ISSUE #156: Custom Edit Route when "use default" is disabled fails to check if form mode is enabled
# What is this?

This is a bug/not well tested functionality

We introduced a custom "EDIT" tab to allow users to have the same "feel" when the "default" edit form is disabled for them but other form Modes are enabled. 

The whole functionality is here:
https://github.com/esmero/strawberryfield/blob/main/src/Controller/CustomNodeEditController.php

The problem with this controller is 
1.- It does not verify if a Form mode, for a certain Bundle is or Not active. It only checks for Permissions. And Permissions can differ from what is enabled or not

How to check its bad

Take the demo user
1.- Remove the "Use Default form" permission from it
2.- Only allow the RAW JSON one (e.g make it only for "authenticated users")
3.- On Structure, Content types, press "Manage Form Modes" under the "Digital Objects" drop down menu (right)
4.- Disable "RAW JSON" from the available Form modes.
Save

Log in as Demo, try to press "edit" on any of your own Objects. You will get a Permission denied. OR even worst, if you try to access the actual route, it will fail with ROUTE not found.

This is the method that is sloppy

https://github.com/esmero/strawberryfield/blob/5c2e201bdcf2d11bdf70c9a2fb321413974a3172/src/Controller/CustomNodeEditController.php#L58-L70

pcambra:
This PR makes use of getFormModeOptionsByBundle among other cleanup tasks

Resolves #156

### DIFF
--- a/src/Controller/CustomNodeEditController.php
+++ b/src/Controller/CustomNodeEditController.php
@@ -4,40 +4,33 @@ namespace Drupal\strawberryfield\Controller;
 
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Controller\ControllerBase;
-use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Entity\EntityDisplayRepositoryInterface;
 use Drupal\node\NodeInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
+/**
+ * Class CustomNodeEditController.
+ *
+ * @package strawberryfield
+ */
 class CustomNodeEditController extends ControllerBase {
 
   /**
    * The EntityDisplayRepository service.
    *
-   * @var \Drupal\core\Entity\EntityDisplayRepository $entityDisplayRepository
-   *
+   * @var \Drupal\core\Entity\EntityDisplayRepository
    */
   protected $entityDisplayRepository;
 
   /**
-   * Node form modes active for this current user
-   *
-   * @var array
-   *
-   */
-  protected $activeNodeFormModes;
-  
-  /**
-   * Constructs a new CustomNodeEditLocalTasks
+   * Constructs a new CustomNodeEditLocalTasks.
    *
    * @param \Drupal\Core\Entity\EntityDisplayRepositoryInterface $entity_display_repository
-   *
+   *   The EntityDisplayRepository service.
    */
   public function __construct(EntityDisplayRepositoryInterface $entity_display_repository) {
     $this->entityDisplayRepository = $entity_display_repository;
-    $this->activeNodeFormModes = $this->getActiveNodeFormModes();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/Controller/CustomNodeEditController.php
```diff
@@ -4,40 +4,33 @@
...
-   * Constructs a new CustomNodeEditLocalTasks
+   * Constructs a new CustomNodeEditLocalTasks.
    *
    * @param \Drupal\Core\Entity\EntityDisplayRepositoryInterface $entity_display_repository
-   *
+   *   The EntityDisplayRepository service.
    */
   public function __construct(EntityDisplayRepositoryInterface $entity_display_repository) {
     $this->entityDisplayRepository = $entity_display_repository;
-    $this->activeNodeFormModes = $this->getActiveNodeFormModes();
```
DiegoPino, line 40: I may fight this one but open to listen to arguments! With this change we call now ::getActiveNodeFormModes twice and each one calls again the list from storage (I know it may be cached but cached is still DB) and does the array_filter. Having it saved as protected implies we do not need to make the call again and filter again but also that in a "very very edge" case where the active ones change during (yes, I'm crazy) call to the controller, our initial assessment of what is available will be consistent with our checks/redirect?

src/Controller/CustomNodeEditController.php
```diff
@@ -50,60 +43,50 @@ public static function create(ContainerInterface $container) {
...
-      if ($this->currentUser()->hasPermission("use {$value['id']} form mode")) {
-        $activeFormModes[$key] = $value;
-      }
-    }
-    
-    return $activeFormModes;
+  public function getActiveNodeFormModes(NodeInterface $node) {
+    $formModes = $this->entityDisplayRepository->getFormModeOptionsByBundle('node', $node->bundle());
+    return array_filter($formModes, function ($mode) {
+      return $this->currentUser()->hasPermission("use node.$mode form mode");
```
DiegoPino, line 54: Nice. Let's use {$mode} to be consistent with {$firstActiveFormMode}. It is just cosmetics I know. This is a side note, I will in general (and I will make exceptions) for expected large Arrays prefer an foreach instead of an array filter (for performance) but in this case filter reads better and the array is always expected to be small, so good, good!

src/Controller/CustomNodeEditController.php
```diff
@@ -50,60 +43,50 @@ public static function create(ContainerInterface $container) {
...
-  * Grants access if user has access to a named form mode, but not default form mode.
-  *
-  * @param \Drupal\Core\Session\AccountInterface $account
-  *   Run access checks for this account.
-  *
-  * @return \Drupal\Core\Access\AccessResultInterface
-  *   The access result.
-  */
-  public function access(AccountInterface $account) {
-    $accessDefaultFormMode = $account->hasPermission('use default form mode');
```
DiegoPino, line 83: So was this also a bug here? `use default form mode` ? instead of  "use $entity_type_id.default form mode"? gosh!!!! Thanks!

DiegoPino (commented):
Comments here. Won't go Request Changes because I would love to hear your opinion on the first comment!

src/Controller/CustomNodeEditController.php
pcambra: This is what I thought too but on my tests it is only called once as I think is statically cached by default, it would be good to double check

src/Controller/CustomNodeEditController.php
pcambra: Oh I am a big fan of closures :)

src/Controller/CustomNodeEditController.php
DiegoPino: Oh, that is confusing? So you say the whole controller gets all its methods statically cached? That is a bit obscure. Should I test too?

src/Controller/CustomNodeEditController.php
pcambra: I was testing it and I must confess I don't fully understand the access method but it does seem to be  $entity_type_id.$mode, according to FormModeManagerBase::setUpUsers

src/Controller/CustomNodeEditController.php
pcambra: I would appreciate if you could test, in my debug sessions, it was only being called once, as I was considering adding a static. 
Otherwise we need to use a different method such as getFormModeOptions as I don't think we have the node context on construct or at least I don't know how to get it without ugly routematches calls.

DiegoPino: Ok, that makes sense. Will move forward with this the way it is.

Issue #171, opened 2021-04-03
Add an Empty MiniOCR as a Constant to Strawberry Flavor Data Source


DiegoPino:
This is needed for SBF and for the new OCR Solr plugin (version > 0.5) because it now uses a validating XML parser to get the coordinated and empty pages do not generate that.

DiegoPino:
@giancarlobi falling asleep here. I have the code for both modules but won't be able to commit now. Will do first hour in the morning, I promise

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/88de895dd7ea8b3452a1135e8f60ea1fe4f387e7

Issue #170, opened 2021-03-31
Access to strawberryfield_flavor_datasource data disabled for anonymous


giancarlobi:
@DiegoPino In a Views, when we have to access data of strawberryfield_flavor_datasource, I have to config advanced query settings to by-pass permission check if not anonymous can't see view results.

DiegoPino:
Thanks @giancarlobi 
Looking at it: All that magic happens in the Access Control Processor: 
\Drupal\search_api\Plugin\search_api\processor\ContentAccess

I will look at how our `strawberryfield_flavor_datasource` deals (at $item) level with giving the Processor the data it needs to index the right Values. Thanks!

DiegoPino:
@giancarlobi in the meantime. If you disable the "Access Control" Search API Processor, SBF access control is respected correctly. I will build eventually a Processor just for this, but it seems to be that the Access Control processor adds more burden. Why? Because without it, by just changing the Status (e.g to Archived) to an ADO that generated HOCR, SBF respects that and stops appearing immediately from Views without reindexing. With the Processor (and if I fix it) it would require a reindex.

Only benefit I see of the processor is that it would return "less" Documents for the same query. Give it a try if you have time

giancarlobi:
@DiegoPino Friend, if I have some time this weekend I'll check this, to be clear, is this the UI where I have to disable "Access Control" Search API Processor?
![image](https://user-images.githubusercontent.com/4232178/113479498-c28a2680-948f-11eb-9647-1a5f2ec02f81.png)

DiegoPino:
@giancarlobi sorry for the delay, got distracted with life and code (AMI). Yes, that is the checkbox. Once you disable, reindex. The behavior I saw is that Views will respect Entity Access and since SBF are exposed as "nodes" even if they are not, if an ADO that produced them can not be seen by a certain user (because its Archived, draft or Access Control does not allow that user to see it), also OCR (pages) will be hidden.

Please give it a look and let me know if you see the same behavior.
The only down site (means I will eventually build that processor, I promise) is that the query itself will generate results (all the ones that would match) and it is Drupal Views and core Access Control that filters them out. That may have performance issues, but since there is also paging happening, maybe not that bad? (at least until I code the whole thing).

Thanks a lot and again sorry. Coding as fast as I can!

giancarlobi:
@DiegoPino On one archipelago that works fine:
- I restored default views query settings (unflagged by pass access control)
- I disabled access control processor as above image
[...]

PR #172, opened 2021-04-04
ISSUE-171: Adds en empty miniOCR as constant and ensure Access Control

DiegoPino:
See #171 and #170 

This adds a constant for empty miniOCRs that can be used by runners or any other (e.g an XML OCR endpoint) and also ensures that we do Access Control on Parent Node.

And extra Search API query option was added (`no_highlight`) to help in cases where we want to programatically call a Solr query, get all Fulltext fields but do not want HL enabled, because without an exact Fulltext field ocrhighlight, as documented, will fail with an "invalid offset" error Solr side.

@giancarlobi you probably need this to test #170

### DIFF
--- a/src/Plugin/search_api/datasource/StrawberryfieldFlavorDatasource.php
+++ b/src/Plugin/search_api/datasource/StrawberryfieldFlavorDatasource.php
@@ -2,8 +2,10 @@
 
 namespace Drupal\strawberryfield\Plugin\search_api\datasource;
 
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityDisplayRepositoryInterface;
+use Drupal\Core\Session\AccountInterface;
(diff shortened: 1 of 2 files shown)


DiegoPino:
This requires https://github.com/esmero/strawberry_runners/pull/35 to test empty pages under  versions higher than https://github.com/dbmdz/solr-ocrhighlighting/releases/tag/0.5.0 

@giancarlobi i will try to keep up with future compatibility but can't ensure I can test for unreleased tags (sorry to many moving targets), so if you can help with that it would be great (and very appreciated). We will for now keep releasing Archipelago with 0.5.0

DiegoPino:
@giancarlobi let me know if this fixes for you empty HOCR. Once you are OK I will merge this and SBR counterpart. Thanks a lot

giancarlobi:
@DiegoPino It works fine, also with solr-ocrhighlighting-0.6.0-SNAPSHOT.jar so for me ok to merge, friend

PR #176, opened 2021-04-30
1.0.0 rc2 textselect

giancarlobi:
@DiegoPino This is just the first step to achieve (maybe) text select option on IAB.
I try to be backward compatible, i.e. added two new parameters (format and page) in the route with default values so original route for IAB search results works without any change.
I opened this pull linked to a brach on my git where I'll commit next code.


giancarlobi:
@DiegoPino we need laso https://github.com/esmero/format_strawberryfield/pull/154 to make all working.

src/Controller/StrawberryfieldFlavorDatasourceSearchController.php
```diff
@@ -129,6 +134,39 @@ public function search(Request $request, ContentEntityInterface $node, string $f
...
+      $response = new CacheableResponse(
+        $output,
+        200,
+        ['content-type' => 'application/xml'],
+        TRUE
+      );
+      if ($response) {
+        // Set CORS. IIIF and others will assume this is true.
+        $response->headers->set('access-control-allow-origin', '*');
+        $response->addCacheableDependency($node);
```
DiegoPino, line 163: I need to check if we can add a DataSource ID as cacheable response too. That way if the Page itself changes/is added we get the right way. Also we should not cache empty outputs or not found pages

strawberryfield.routing.yml
```diff
@@ -74,9 +76,10 @@ strawberryfield.flavor_datasource_search:
       resource_type:
         type: 'ado'
       fileuuid: 'all'
+      processor: 'ocr'
+      format: 'json'
+      page: 'all'
   requirements:
-    fileuuid: .+
```
DiegoPino, line 78: WE still need the fileuuid? and the processor?

giancarlobi: I think we don't need fileuuid anymore while processor (currently only ocr) could be useful for future flavors, right?

DiegoPino: I think we still need it. Because page itself is not enough when dealing with TIFFs instead of PDFs

giancarlobi: @DiegoPino I had to change routing yml as you see to make default working  because I want it works also without passing format and page to be backward compatible. Some test give me that, removing those settings from requirements while they are present under parameters. Probably you have more clear idea about this, feel free to make any changes you consider better for code

DiegoPino: I know this may be more work(one I can do) but maybe we want to have the XML format (since it won't really search right? it is the full thing?) into its own route then. I feel unsure allowing so much unchecked arguments.Give me a few, I have some fire to extinguish on another project. Thanks!!

giancarlobi: @DiegoPino Full miniocr as stored in Solr is already returned if you use /miniocr/page as in this https://archipelago.byterfly.eu/do/7facaffe-e07d-4931-a5e3-0dc8210d7518/flavorsearch/all/ocr/miniocr/10 is this what you mean? anyway, no rush, friend

DiegoPino:
Closing. Resolved via https://github.com/esmero/strawberryfield/commit/53ab33258533a4553de876f4fb227a4de92c729e

Issue #175, opened 2021-04-28
EDTF Date Strawberryfield Key Name provider


patdunlavey:
Need a way to index EDTF date strings as dates in solr. This document captures the issues/complexity, I think: https://lyrasis.zendesk.com/hc/en-us/articles/360041734211-Deep-dive-Best-practices-for-dates-in-Islandora-MODS-metadata

@DiegoPino I could not find another github issue for this, but if there is, feel free to close this one and let me know where it is. I'm willing to have this task assigned to me if I can grab some time with you to get me pointed in the right direction.

DiegoPino:
@patdunlavey good. There is no other one specific to this one open (there is a global mention of a date aware one but this is better).

My guts here:
1.- ISO8601 is a subset of ETDF so we need only a single Key Name Provider and the key name provider does not need to know the source/format selected.
2.- Key Name providers are optimistic and opportunistic. They will try to provide data in the chosen TypeData format but if that is not possible they should return empty (not failure).
3.- Key Name provider should be Solr Date Range aware and provide the proper format. But if so, then the output should be only for Ranges since the Solr Field assigned at the Search API level needs to match one type only. Means if you need point and ranged dates you need to configure two key name providers with different settings. Means also parsing ranges/points needs to be a setting
4.- Human readable expression should be also an option. If selected the output format will be string. 
5.- You can also test outputting always strings, just with the right format. Search API should be able to cast them if the format is right to a date (would mean a less complex Key Name provider too)

Hope this helps. I would use the https://github.com/esmero/strawberryfield/blob/1.0.0-RC2/src/Plugin/StrawberryfieldKeyNameProvider/JmesPathNameProvider.php or even extend it as base class. Since it includes already the "get value via JMESPATH and get multiple JMESPATHS as a single one" functionality

DiegoPino:
Solved vy https://github.com/esmero/strawberryfield/commit/79f30cd791c569954c299a4cc158d4f8c2b0580d and https://github.com/esmero/strawberryfield/commit/13a20d31a12a6c79bf800bd9502f467667ca3af2

The JmesPath provider now supports most Date strings, including ISO8601, free text cast-able to DateTime and EDTF individual and range formats. The key name provider has a checkbox that enables this logic. If the value does not match the returning value will be NULL

PR #182, opened 2021-05-15
ISSUE-175: First pass (working) on EDTF/ISO8601 validation at Key Name Provider …

DiegoPino:
…level
See #175 

@dmer @giancarlobi @patdunlavey, this got me stuck during the RC2. Needed to facet on Dates and for some reason a CSV in the new AMI demo Set had a "full description" inside a  date JSON key in the metadata and Solr Indexing was breaking (bad). 

So i added a checkbox to the `JmesPath`  Key Name provider  (easier than recoding it) and implemented, if that is enabled, (FALSE by default) ETDF/ISO8601 parsing, max and min extraction and indexing back. So if you add a date like 2015-02-02/2020-03-03 it will add both to the facets and index. If the value is totally non-OK and the checkbox is enabled and you Set your Solr to Date for the field, all good. 2015-XX-XX? Also! etc, etc, etc.
Of course if you do not check the box and then force Solr to a date field, it will fail with strings like "the best time of my life" which is a future wish, not a date!

This was the missing piece for the RC2. Now i can sleep

### DIFF
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -50,7 +50,10 @@ strawberryfield.strawberry_keynameprovider.jmespath:
   mapping:
     source_key:
       type: string
-      label: 'A Comma separated string containing one or more JMESPaths '
+      label: 'A Comma separated string containing one or more JMESPaths'
+    is_date:
+      type: boolean
+      label: 'If the value should be considered and validated as a date'
     exposed_key:
       type: string
       label: 'The field property we expose for the Strawberryfield'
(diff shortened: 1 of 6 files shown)


DiegoPino:
Also I'm falling asleep and have tested this quite deeply for like 3 hours. So merging. Time is precious

patdunlavey:
@DiegoPino this is fantastic, thank you so much. I look forward to checking this out.

DiegoPino:
@patdunlavey thanks! We will also need to code a complete date facet widget. The one include in Drupal simple does not work. Its full of bugs. And we may now improve this one by adding also more features like the humanizer too. Hugs

Issue #184, opened 2021-05-22
Add extra fields to Strawberry Flavors Data Source


DiegoPino:
# What? A few extras Data properties so our Solr Documents for HOCR/WACZ URL/TEXT etc are richer.

- A Label for each Docs,ent. For Book pages having a Sequence 1, 2, 3 is enough. If we are using this for other things like processed data from WARC/WACZ file we may want to have the URL.
- Metadata. Not a string, but a list of strings. For now this may be a generic single one but I want to test using NLP to extract entities (people, places, etc) here and push those as part of the Solr Document.

These changes will only need to be applied here for now. I will make sure that they are "optional" and default to something in case the Processors do not provide them yet.

@giancarlobi this is your interest area of course!

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/1d1cff16f3bde4d6da06a156f8e9306551388e87

PR #185, opened 2021-05-22
ISSUE-184: Adds new data properties to Class StrawberryfieldFlavorDatasource and provide custom Access Control

DiegoPino:
See #184 

`Label` and `Metadata`, `ts`, `url`, `sentiment`, `who`, `where` and `when` are the new ones. Will come handy in the future (near future)

Will test if original functionality is not affected and will go from there. This is the first step into RC3 and will also include once tested the new reactPHP code Giancarlo did to enhance the experience

Also includes Access Control processor (replacement for the Core one) that knows how to handle Strawberry Flavors (so full status connection between Nodes and its HOCR) and some extra goodies like automatic updates on Strawberry Flavor Solr Indexed Documents(HOCR when the Node changes user/status).

This comes with an extra pull from Strawberry Runners but does not require it.


giancarlobi:
@DiegoPino I merged this into my RC2-textselection and all seems work fine. I added a new ADO, added new fields to flavor solr datasource index, reindexing, all good, no errors.

PR #187, opened 2021-06-03
1.0.0 RC3 Text Select and Alto

giancarlobi:
@DiegoPino WIP Text selection for IAB + ALTO


DiegoPino:
@giancarlobi will merge first into ISSUE-201 to give this a last Drupal Coding Standards Pass. Sorry for the delay, this is priority number one but since I'm slow and there is a lot of code to review i kept pushing this. By merging this way we keep all your commits and of course attributions to this work. Thanks!

Issue #188, opened 2021-06-11
Use Mediainfo for Video/Audio Tech metadata Extraction


DiegoPino:
# What?

We need duration/codecs/metadata/etc from Media (Video/Audio) that exif nor pronom can provide. So we go for mediainfo

## Tasks:
- Add mediainfo to the PHP Docker container (DONE! esmero/php-7.4-fpm:1.0.0-RC2-multiarch contains it now)
- Include a Mediainfo PHP library so we do not need to deal with filesystem level parsing of output
- Add a configuration option in this module (same place we use for pdfinfo, etc)
- Write the wrapper code that extracts Media Info and pushes it into JSON

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/494d522f2dce79c202e533b72582c8b1829eb9af

PR #189, opened 2021-06-11
ISSUE-188: Media info/techmedata for Video/Audio and also EXIF filters

DiegoPino:
See #188 

For media Info: This is the current full output. We may filter/slim this down a bit
This pull now depends on https://github.com/esmero/format_strawberryfield/pull/169, both need to be merged together.

- https://gist.github.com/DiegoPino/aa8a90edc8c136ca1ac9555eccf76e72

~Since I'm here I also added traditional strtostring processing for the JMESPATH Key provider when is date is checked. It seemed like a good idea/attempt at least to get some dates in~ NOTE: this was a good attempt but was improved a lot in https://github.com/esmero/strawberryfield/pull/194 and already merged 

- Includes Revision persistence as files (with options)
- New Key Name provider overview tab (visit /admin/structure/strawberry_keynameprovider)
Screen Shot 2021-08-29 at 10 13 16 PM


DiegoPino:
Ok, refactor is looking better. I just need to add the new object path for the revision based DO deposit

Issue #196, opened 2021-09-14
Make digital object storage folder configurable


patdunlavey:
# What

The file persister destination folder for digital object files is hard-coded in [StrawberryfieldEventInsertSubscriberDepositDO->onEntityInsert()](https://github.com/esmero/strawberryfield/blob/main/src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php#L122) to `://dostorage`. I don't see any place where this can be altered.

In our case, we want to store digital object files for staging and production into separate folders while still using the same s3 bucket. 

# Possible solutions
1. Modify the path construction logic so that the `://` is stored as part of the `strawberryfield.storage_settings.object_file_scheme`. This would require a change in the code that stores the object_file_scheme, as well as the code that reads it. Then a form alter could be written targeting the FilePersisterServiceSettingsForm storage schema options, providing different options for staging and production.
2. Add an alter hook immediately or soon after the above referenced line where $path is defined. I'll note that an alter hook is provided for `StrawberryfieldFilePersisterService::getDestinationUri`, so consistency would argue for this as the preferred option. 
3. Actually make the storage folder configurable on the file persistence settings form. Add a text field labeled "Storage Path for Persisting Digital Objects", with the default value "dostorage", but where one could enter e.g. "production/dostorage".

@DiegoPino, I'll be happy to send you a PR for whatever solution you think makes most sense.

DiegoPino:
@patdunlavey yes, it can not be altered right now. I think keeping the scheme around as it is (so not adding the ://) + adding a secondary "folder/prefix" option/form element/setting may be Ok. And either a hook (which may all disappear by Drupal 10) or Emmitt a Symfony event other modules can use (with a subscriber) to alter the $path too. The event is the preferred way and SBF will eventually get refactored to use an event too (no worries, won't remove the alter hook) for modifying the file persistence part.

Happy to review a Pull request (which reminds me please check your code review on the XDEBUG documentation which has unanswered change requests coming from the team). 

For the pull please check https://github.com/esmero/strawberryfield/pull/189/files since there is also a new mode there which adds a new Eventsubscriber for the same purpose. Thanks!

patdunlavey:
I need to learn the event subscriber system, so I'll give that a whirl. Do you not like the third option (just provide configurability of the folder path(s) on the form, and don't require a developer intervention (hook or even subscriber)?

DiegoPino:
[...]

PR #198, opened 2021-09-15
ISSUE-196-custom-file-storage-path --- provide options to set relative paths for storage of managed strawberryfield files and digital object files (json)

patdunlavey:
See #196 

This PR provides configuration options on the persistent file storage settings page to set relative paths for storage of the strawberryfield uploaded files (default none) and the digital object files (json) (default "dostorage").


patdunlavey:
@DiegoPino let me know your thoughts. Thanks!

config/schema/strawberryfield.schema.yml
```diff
@@ -79,9 +79,17 @@ strawberryfield.storage_settings:
     file_scheme:
       type: string
       label: 'Storage Scheme for Persisting Files'
+    file_path:
```
DiegoPino, line 82: Ok, do we need both? if we add here the file_path then the alter hook needs to be adapted again? Just wondering since the ISSUE was about the Object File Path. But happy to deal with both (we need an alter too in the form)

src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php
```diff
@@ -91,6 +98,9 @@ public function __construct(
     $this->destinationScheme = $config_factory->get(
       'strawberryfield.storage_settings'
     )->get('object_file_scheme');
+    $this->destinationRelativePath = $config_factory->get(
+      'strawberryfield.storage_settings'
+    )->get('object_file_path') ?? "dostorage";
```
DiegoPino, line 103: This looks good. What if you make `dostorage` a class constant?

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -47,7 +47,12 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#required' => TRUE
 
     ];
-
+    $form['file_path'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Relative Path for Persisting Files'),
+      '#description' => $this->t('Path relative to the root of the storage scheme selected above where the hashed directories used to store persisted managed files will be created. Do not include beginning or ending slashes. Default is "".'),
```
DiegoPino, line 53: Let's add a warning here. That says "changing this path after your ingests won't modify existing  Files"

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -47,7 +47,12 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#default_value' => !empty($config_storage->get('file_path')) ? $config_storage->get('file_path'): "",
```
DiegoPino, line 54: Wonder if we should better go for NULL instead of empties? And then change the logic of the setting? Not sure really

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -306,9 +317,11 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       ->save();
     $this->config('strawberryfield.storage_settings')
       ->set('file_scheme', $form_state->getValue('file_scheme'))
+      ->set('file_path', trim($form_state->getValue('file_path')," \n\r\t\v\0/"))
```
DiegoPino, line 320: Lovely! But I think for this case we need to make sure these paths are sanitized. Also S3 has size limits. Look at the method where I refactored the hook. I have a few notes there about this

src/StrawberryfieldFilePersisterService.php
```diff
@@ -223,7 +231,10 @@ public function getDestinationUri(
   ) {
 
     // Default $relativefolder is a 3 char hash generated by a checksum algorithm.
-    $relativefolder = substr($checksum, 0, 3);
+    // First, get any parent directories.
+    $relativefolders = explode('/', $this->storageConfig->get('file_path') ?? "");
+    $relativefolders[] = substr($checksum, 0, 3);
+    $relativefolder = implode('/', array_filter($relativefolders));
```
DiegoPino, line 237: Ok, this needs to be added to the refactor too

DiegoPino:
@patdunlavey thanks for the pull. I added a few questions really more than code reviews. Looks good, i would mostly like to do some more sanitization on the inputs and make sure there is no way a user can corrupt storage via this forms. This also means I need to refactor my #199 to match the new paths.

config/schema/strawberryfield.schema.yml
patdunlavey: Yes, this is not strictly necessary since we can modify the path through hook_strawberryfield_file_destination_alter(). However, it seems to me that being able to set the path would be desired pretty often, and requiring a user to implement a hook just to set this to a different path seems a bit onerous, particularly as they will see that we do give them the configuration option for the digital object file storage.
I don't feel strongly about this though, Happy to take it out if you think that makes more sense.

src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php
patdunlavey: Done.

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: Ok, I added this note to both fields. Tell me if that's not correct!

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: My thinking was that the user input for this field is trimmed, exploded on slashes, etc - all operations that expect a string.

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: So checkPlain and limit the path to 128 characters? Ok to just set the field character limit?

src/StrawberryfieldFilePersisterService.php
patdunlavey: I'm thinking let's get your refactor merged first, and then resolve this branch it. Does that make sense?

DiegoPino: Ok, let's roll with that. Will wait for your review on that pull before merging it. Thx

src/Form/FilePersisterServiceSettingsForm.php
DiegoPino: I think we need to make folder validation method. Should not have spaces. Should have a controlled list of characters (like no utf8 strawberries).  Here is a list of what needs to be checked. I agree we are also allowing local storage, e.g what @giancarlobi does, but still not sure I want to have on a local unix filesystem something that fights a 1986 unix spec too much. See https://www.ezs3.com/public/232.cfm

src/Form/FilePersisterServiceSettingsForm.php
DiegoPino: Ok, let's roll with an empty string

src/Form/FilePersisterServiceSettingsForm.php
DiegoPino: Thanks. Correct!

config/schema/strawberryfield.schema.yml
DiegoPino: Ok, let's keep it. I understand how this can be useful for vanilla implementations!

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: Ok, I'll work on a validation method.

patdunlavey:
Update: #199 has been merged. I need to address the file path sanitation issues (in the next day or so) and then will ask for further review.

patdunlavey:
I've added validation for the two file path fields - both ajax and upon submit. It's probably overkill, but it's working pretty good for me. The one thing that had me scratching my head for a good while was how to validate the system file path (public and private schemes). As I commented in the code, neither \Drupal::service('path.validator')->isValid() nor \Drupal::service('stream_wrapper_manager')->isValidUri() work --- at all! Well, I probably don't understand how they're supposed to work. I ended up using FileSystem::prepareDirectory() to test if the directory path provided could be written to. Can you review that part in particular, @DiegoPino ? There is almost certainly a better way to do it!

patdunlavey:
Hi @DiegoPino. With all the discussion of the CSV exporter today I don't want to lose sight of this issue being the only one we know of that is actually blocking us from starting to ingest content into our production instance. Any chance you could take a bit of time to give this a look?

DiegoPino:
@patdunlavey I can tomorrow. Did you adapt the code so it merges clean with the last hook change? This will break my open https://github.com/esmero/strawberryfield/pull/189 so give me at least 24 hours, hands totally full. Will check back when ready. thx

patdunlavey:
@DiegoPino yes it merges clean with the last hook change (that change has been merged to 1.0.0-RC3 and this PR is against that). I'll take a look at #189. Maybe I can create a combined merge branch? Or if that branch is ready to merge, maybe you could do that and then I can reconcile again with this branch?

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -56,6 +73,22 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#title' => $this->t('Relative Path for Persisting Digital Object Files'),
+      '#description' => $this->t('Path relative to the root of the storage scheme selected above where digital object files will be stored. Do not include beginning or ending slashes. Default is "@storage".
+                                  <br>Note that changing this setting will not affect the file storage locations for previously ingested objects. They will remain where they were.',
+        ['@storage' => StrawberryfieldEventInsertSubscriberDepositDO::DEFAULTOBJECTSTORAGEFILEPATH]),
+      '#default_value' => !empty($config_storage->get('object_file_path')) ? $config_storage->get('object_file_path') : StrawberryfieldEventInsertSubscriberDepositDO::DEFAULTOBJECTSTORAGEFILEPATH,
+      '#prefix' => '<span class="filepath-exec-path-validation"></span>',
+      '#ajax' => [
+        'callback' => [$this, 'validateObjectFilePath'],
+        'effect' => 'fade',
+        'wrapper' => 'objectfilepath-exec-path-validation',
```
DiegoPino, line 87: The `exec` is a bit confusing. Better remove since there is no execution at all

src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php
```diff
@@ -91,6 +103,9 @@ public function __construct(
     $this->destinationScheme = $config_factory->get(
       'strawberryfield.storage_settings'
     )->get('object_file_scheme');
+    $this->destinationRelativePath = $config_factory->get(
```
DiegoPino, line 106: Like this

src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php
```diff
@@ -20,6 +20,11 @@ class StrawberryfieldEventInsertSubscriberDepositDO extends StrawberryfieldEvent
 
   use StringTranslationTrait;
 
+  /**
+   * Default path for digital object file storage.
+   */
+  public const DEFAULTOBJECTSTORAGEFILEPATH = 'dostorage';
```
DiegoPino, line 26: really like this. Could you add a few "_" to make it more legible? Sorry. It really helps

config/schema/strawberryfield.schema.yml
```diff
@@ -79,9 +79,17 @@ strawberryfield.storage_settings:
...
+      type: string
+      label: 'Storage Path under the Storage Scheme, for Persisting Files'
+      default: ""
     object_file_scheme:
       type: string
       label: 'Storage Scheme for Persisting Digital Objects in JSON format'
+    object_file_path:
+      type: string
+      label: 'Storage Path under the Storage Scheme, for Persisting Digital Objects in JSON format'
+      default: "dostorage"
```
DiegoPino, line 92: good!

src/StrawberryfieldUtilityService.php
```diff
@@ -28,6 +28,20 @@ class StrawberryfieldUtilityService implements StrawberryfieldUtilityServiceInte
 
   use StringTranslationTrait;
 
+  /**
+   * File system path pattern. Must begin and end with a word character, may have
+   * forward slashes delimiting directory names within the path.
+   */
+  const PATHPATTERN = "/(^(\w)+$)|(^(\w)([\/\w-]*)(\w)$)/";
```
DiegoPino, line 35: This is nice. Could we make this one more explicit? VALID_FILESYSTEM_PATH_PATTERN?

src/StrawberryfieldUtilityService.php
```diff
@@ -28,6 +28,20 @@ class StrawberryfieldUtilityService implements StrawberryfieldUtilityServiceInte
...
+
+  /**
+   * S3 file system path pattern.
+   * See \Drupal\strawberryfield\StrawberryfieldUtilityServiceInterface::s3FilePathIsValid
+   * for explanation of the rules.
+   */
+  const S3PATHPATTERN = "/^[a-z0-9][a-z0-9\-]{1,61}[a-z0-9]$/";
```
DiegoPino, line 42: Same here (also super nice)  VALID_S3_PREFIX_PATTERN? (since in S3 paths are normally known as prefixes)

DiegoPino (changes requested):
@patdunlavey this is pretty much in good shape but still needs a bit more work. There is a bit of redundancy in the code (validation methods can be a single) one and S3 v/s any other are handled differently when it does not need to. Some CONST (thanks for that) need spacing (_) to be readable and a few minor things. I can help you tomorrow to get those done and we can merge. Thanks again for working on this.

src/StrawberryfieldUtilityService.php
```diff
@@ -329,4 +343,57 @@ public function getCountByProcessorInSolr(EntityInterface $entity, string $proce
...
+    $test_path = $scheme . "://" . $path;
+
+    // FileSystemInterface::prepareDirectory with the MODIFY_PERMISSIONS flag checks to see if it's present and able to be written to.
+    $valid = \Drupal::service('file_system')->prepareDirectory($test_path, FileSystemInterface::MODIFY_PERMISSIONS);
+    if(!$valid) {
+      // If not present or able to be written to, let's check to see if we can create it.
+      $valid = \Drupal::service('file_system')->prepareDirectory($test_path, FileSystemInterface::CREATE_DIRECTORY);
+      if($valid) {
+        // Successfully created the directory, so we delete it now.
+        \Drupal::service('file_system')->rmdir($test_path);
```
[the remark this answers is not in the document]
patdunlavey: Thanks for the dependency injection correction - this symfony stuff is still very foggy to me!
I'm with you that creating a directory and then deleting it seems dangerous. I was hoping you could point me to another way to determine if the proposed destination path is writable!
I'll leave the two-step logic as-is unless and until we can think of something more elegant.

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: Yep, a lazy copy-paste mistake.

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -145,10 +178,155 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+  ) {
+    $response = new AjaxResponse();
+    $path = $form_state->getValue('file_path');
+    $scheme = $form_state->getValue('file_scheme');
+    if (empty($path)) {
+      $valid = TRUE;
+    }
+    else {
+      switch ($scheme) {
+        case 's3':
```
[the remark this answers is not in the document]
patdunlavey: I'm nervous about removing the regex tests because I'm not 100% sure I like the invasive method of determining writability. Am I just being silly?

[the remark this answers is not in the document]
DiegoPino: Oh, do not remove them. What I say here is that you can deal with any wrapper in the same way. and to be honest, to allow portability between filesystem, the MOST restrictive regex should be the one that is applied to both (so the S3). That makes also your code simpler @patdunlavey . That way moving from local filesystem to the cloud is seamless. If not you have to update all your ADOS!

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -145,10 +178,155 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+            ->s3FilePathIsValid($path);
+          break;
+        default:
+          $valid = \Drupal::service('strawberryfield.utility')
+            ->internalFilePathIsValid($scheme, $path);
+          break;
+      }
+    }
+    if (!$valid) {
+      $warning_message = self::filePathErrorMessage('object file', $scheme,
```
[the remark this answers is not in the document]
patdunlavey: Fixed.

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -145,10 +178,155 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+            ->s3FilePathIsValid($path);
+          break;
+        default:
+          $valid = \Drupal::service('strawberryfield.utility')
+            ->internalFilePathIsValid($scheme, $path);
+          break;
+      }
+    }
+    if (!$valid) {
+      $warning_message = self::filePathErrorMessage('file', $scheme, $path);
```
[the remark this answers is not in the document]
patdunlavey: Fixed.

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -145,10 +178,155 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+   * Utility function builds warning message string for file path validation
+   * functions.
+   *
+   * @param $type
+   * @param $scheme
+   * @param $path
+   *
+   * @return mixed
+   */
+  private function filePathErrorMessage($type, $scheme, $path) {
```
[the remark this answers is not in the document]
patdunlavey: See latest commit. Fairly major refactoring of a lot of this stuff since we are changing to using the same validation rule for S3 and local file systems.

src/StrawberryfieldUtilityService.php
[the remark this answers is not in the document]
patdunlavey: Changes committed. Please re-review.

src/Form/FilePersisterServiceSettingsForm.php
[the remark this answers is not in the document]
patdunlavey: Ok, switched to using the S3 prefix pattern for all. Please re-review!

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: Fixed.

src/EventSubscriber/StrawberryfieldEventInsertSubscriberDepositDO.php
patdunlavey: Better now, I hope!

src/StrawberryfieldUtilityService.php
patdunlavey: Better now, I hope.

src/StrawberryfieldUtilityService.php
patdunlavey: Better now I hope.

patdunlavey:
I think I have addressed all your review comments. The main thing was refactoring to only use one validation method (S3). Got rid of a fair bit of redundant code. Made the error message even more wordy! Ready for your re-review when you get a chance. Thanks so much @DiegoPino , you are amazing!

DiegoPino:
@patdunlavey amazing! Will give it a spin at 2:30 and will let you know. Thanks!!

src/Form/FilePersisterServiceSettingsForm.php
```diff
@@ -471,15 +426,14 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
-        switch ($scheme) {
-          case 's3':
-            $valid = \Drupal::service('strawberryfield.utility')
-              ->s3FilePathIsValid($path);
-            break;
-          default:
-            $valid = \Drupal::service('strawberryfield.utility')
-              ->internalFilePathIsValid($scheme, $path);
-            break;
+        // Validate for known schemes.
```
DiegoPino, line 429: @patdunlavey this looks much better. I will leave this at it is but will refactor tonight a bit. Why?

We have this here:
 $scheme_options = OcflHelper::getVisibleStreamWrappers();

If a scheme is not in that list, then it is actually invalid and we need to validate as "false". No worries. It works the way you wrote but I feel its good we reuse code.

src/StrawberryfieldUtilityService.php
```diff
@@ -97,15 +93,22 @@ class StrawberryfieldUtilityService implements StrawberryfieldUtilityServiceInte
...
+   * @param  \Drupal\Core\File\FileSystemInterface  $file_system
+   * @param  \Drupal\Core\Entity\EntityTypeManagerInterface  $entity_type_manager
+   * @param  \Drupal\Core\Config\ConfigFactoryInterface  $config_factory
+   * @param  \Drupal\Core\Extension\ModuleHandlerInterface  $module_handler
+   * @param  \Drupal\Core\Entity\EntityFieldManagerInterface  $entity_field_manager
+   * @param  \Drupal\search_api\ParseMode\ParseModePluginManager  $parse_mode_manager
    *   The Search API parse Manager
    */
-  public function __construct(FileSystemInterface $file_system, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, ModuleHandlerInterface $module_handler, EntityFieldManagerInterface $entity_field_manager, ParseModePluginManager $parse_mode_manager) {
+  public function __construct(
```
DiegoPino, line 104: @patdunlavey I see a lot of DCS changes here. Some super good (thx) other a bit crazy and overly strict. I will also give this file a second pass after the merge to restore some of lost read-ability (but still keeping it DCS)

src/StrawberryfieldUtilityService.php
```diff
@@ -163,25 +167,28 @@ public function getStrawberryfieldMachineNames() {
...
    */
-  public function getStrawberryfieldConfigFromStorage($bundle = 'digital_object') {
+  public function getStrawberryfieldConfigFromStorage($bundle = 'digital_object'
+  ) {
     $field_config = $this->entityTypeManager->getStorage('field_config');
     $field_ids = $field_config->getQuery()
       ->condition('entity_type', 'node')
       ->condition('bundle', $bundle)
-      ->condition('field_type' , 'strawberryfield_field')
+      ->condition('field_type', 'strawberryfield_field')
```
DiegoPino, line 191: nice! gosh.

src/StrawberryfieldUtilityService.php
```diff
@@ -191,24 +198,29 @@ public function getStrawberryfieldConfigFromStorage($bundle = 'digital_object')
   /**
    * {@inheritdoc}
    */
-  public function getStrawberryfieldMachineForBundle($bundle = 'digital_object') {
+  public function getStrawberryfieldMachineForBundle($bundle = 'digital_object'
```
DiegoPino, line 201: @patdunlavey like this one. Totally crazy how PHP Storm reformatted this

src/StrawberryfieldUtilityService.php
```diff
@@ -246,7 +260,7 @@ public function getStrawberryfieldSolrFields(Index $index_entity) {
   /**
    * {@inheritdoc}
    */
-  public function verifyCommand($execpath) :bool {
+  public function verifyCommand($execpath): bool {
```
DiegoPino, line 263: good!

src/StrawberryfieldUtilityService.php
```diff
@@ -346,54 +367,48 @@ public function getCountByProcessorInSolr(EntityInterface $entity, string $proce
   /**
    * {@inheritdoc}
    */
-  public function internalFilePathIsValid(string $scheme, string $path): bool {
+  public function filePathIsValid(string $scheme, string $path): bool {
     // Empty path is always valid.
-    if(empty($path)) {
+    if (empty($path)) {
```
DiegoPino, line 372: @patdunlavey You trimming this somewhere right? Like "       " becomes "" ?

src/StrawberryfieldUtilityService.php
```diff
@@ -346,54 +367,48 @@ public function getCountByProcessorInSolr(EntityInterface $entity, string $proce
...
-    return !empty($valid);
-  }
-
-  /**
-   * {@inheritdoc}
-   */
-  public function s3FilePathIsValid(string $path): bool {
-    // Empty path is always valid.
-    if(empty($path)) {
+    $known_schemes = array_merge(['s3'], array_keys(\Drupal::service('stream_wrapper_manager')->getWrappers(StreamWrapperInterface::LOCAL)));
```
DiegoPino, line 376: Again, we have the OCFL toolbox I wrote to fetch all known schemes that are active and can be used to writing. Will refactor

src/StrawberryfieldUtilityService.php
patdunlavey: Sorry about that. I didn't notice until I had already committed and pushed. If you want, I can reverse these.

src/StrawberryfieldUtilityService.php
```diff
@@ -346,54 +367,48 @@ public function getCountByProcessorInSolr(EntityInterface $entity, string $proce
...
+    $valid = $this->fileSystem->prepareDirectory($test_path, FileSystemInterface::MODIFY_PERMISSIONS);
+    if (!$valid) {
+      // If not present or able to be written to, let's check to see if we can create it.
+      $valid = $this->fileSystem->prepareDirectory($test_path, FileSystemInterface::CREATE_DIRECTORY);
+      if ($valid) {
+        // Successfully created the directory, so we delete it now.
+        $this->fileSystem->rmdir($test_path);
+      }
+    }
+    return !empty($valid);
```
DiegoPino, line 411: @patdunlavey better (for readability) to set $valid to FALSE at the beginning. That way you always return$valid

src/Form/FilePersisterServiceSettingsForm.php
patdunlavey: Makes sense. Thanks!

DiegoPino:
@patdunlavey so I added a few comments. I will merge at it is tonight but will do a small refactor. The DCS pass on the strawberryfield utilities is a bit crazy and there are a few pieces of code that can be simplified by reusing (see same form) existing methods (visible stream wrappers). Thanks. Please do not merge until I refactor after merging. Will mention you. Hugs!

src/StrawberryfieldUtilityService.php
patdunlavey: Ok, thanks for your patience with a D8+ noob!

patdunlavey:
Hugs backatcha!

src/StrawberryfieldUtilityService.php
DiegoPino: No worries. Php Storm does a lot to conform to the 80 lines limit but DCS does not apply that rule for functions. I can do that

DiegoPino (approved):
ready to merge. Small refactor coming later to night! Thanks @patdunlavey on your work on this!

Issue #83, opened 2020-04-30
ACLs and ADO to ADO (Nodes bearing Strawberryfield) Grants inheritance


DiegoPino:
# Good Morning

If you ever had time to read through my [Roadmap feature list](https://github.com/esmero/archipelago-deployment/issues/35) you will have notice probably that i have planned ACL/[access-control-list](https://en.wikipedia.org/wiki/Access-control_list) integration on our Digital Objects in Archipelago. ACL is fundamental for our IR needs to. And it goes way over just nodes, but also files (yeah, let's not speak about media here, since we use file entities).

What Drupal provides with its users and permissions per roles is closer to something named RBAC/[Role based access control](https://en.wikipedia.org/wiki/Access-control_list#Comparing_with_RBAC) but we need more, fain grained, and UI/UX simpler to apply rules but also inheritance (and to be able to control how inheritance works and when it applies) that does not imply a batch operation to copy permissions from parent to child as e.g Islandora 7 does. That is expensive and not in our spirit. 

I read this [good post](https://github.com/Islandora/documentation/issues/413#issuecomment-621493150) from @rosiel today during my coffee and i feel that speaks a little bit about how complex permission system and UI/UX in a hierarchical structure/Drupal can be and what expectations from users differ. I wish i had that type of feedback here! So will humbly borrow that post to get started
# What is  my actual proposal.

Thank to our own 'ahead' planning we have some things in place already to help with this. We lack UI of course because everyone hates to write Forms, but we will get there. I will enumerate my idea

- Every Archipelago Digital Object (means any Node bearing a Strawberryfield) will/can have an ACL.
  - OPTION 1: Attached/logically to the Same SBF but hidden from users. Filtering out keys from our main SBF is easy. 
     - PROS: easier to manage for us, but requires field/wide hiding. Also permissions for the field under the Drupal scheme will apply to ACL itself. Not a big issue since we can swap users on important access operations just to read/access the ACL.  Also File deposit is easier, means 
    - CONS: mixes metadata / and functional access/metadata. Not sure if i like that conceptually. Also, what if some funky use case implies 2+ Strawberryfields? We tell people they don't need it, but we can't tell them what to do. So, in case the ACL is INSIDE a SBF, it would only apply to that SBF. Not sure i like that.
[...]

Issue #68, opened 2020-02-14
Make breadcrumbs semantic aware, flexible and fast


DiegoPino:
# What is that i want?

Everybody knows and loves breadcrumbs. But hey. In semantic, linked data world they are like looking trough a key lock the whole reality. Simple use case everyone knows is Collection/member. But that is not all, E.g, an ADO can be part of two collections, or can be connected to other objects via `isrelatedto`, `partof`, `sameas`, etc, etc. All those things are relevant when thinking about a breadcrumb that is useful. People should be able to check which ones they want in the breadcrumb

So what i want?

Breadcrumbs that read from a list of JSON keys (you define which keys, we could even tag our JMESPATHs that are referencing NODES as such), then fast traverse (i'm quite good at traversing graphs) and accumulate not a single hierarchy but a list of parents. And so on. Direction of the relationship is important. We can do this via direct entityQueries (slow but consistent) or we can use Solr to drive this, fast but requires setting each predicate that people want as Fields, so more config.

UI is tricky. How to draw this tree (its not a cylic graph) in a way that still makes sense to people? Maybe show only one path by default and expand, via JS to more (if there is more) on over any node in this tree? Maybe color coded or prefixed with an Unicode character? having this would be lovely.

marlo-longley:
@DiegoPino I have been thinking about this post!

I really like the functionality and relationships that you describe here. isrelatedto, partof, sameas, multiple parent collections, should all be available information for ADOs and will make Archipelago really rich with info. I am wondering if, trying to force this into the concept of breadcrumbs is too much. I feel that breadcrumbs are known and loved because they are simple and linear, and orient a user to their location in a sitemap. I would love to see a sketch of what you are thinking, because it sounds very cool! I researched existing tree UIs, most are similar to a filesystem/folder browsing, or like an org/flow chart.

On the ADO page itself, to start, all the relationships and collections it is a member of could be listed and hyperlinked in an orderly/pleasing way. Under "Collections" there could be 2 items listed, etc. People can define which relationships they want to show in ADO views like you say. Then we could refine the UI in later steps?

Most people will arrive at an ADO through searching or browsing, and it is the page/view directly prior to reaching the ADO itself where I can more imagine more navigation elements. What do you think?

DiegoPino:
@marlo-longley i agree with the fact that breadcrumbs should be simple/flat. But i also feel, as you say, we need flexible ways of showing the rich complexities of relatationships between ADOs.

This is what i would like:
[...]

PR #205, opened 2021-10-15
ISSUE-68: Breadcrumbs for your ADOs + Other Improvements

DiegoPino:
See #68.

This implementation provides a Config form at `/admin/config/archipelago/breadcrumbs` with two choices.
1. Longest Trail. If 2 more are the same size it will get display the first one.
2. Smart Longest Trail that decides based on the other shorter ones by comparing if amongst the longer ones there are commonalities with the shorter ones. If so, the ones with more "commonalities" are better suited and selected

Also a choice for "disabled/enabling" the breadcrumbs

This will also lead to #83 since the recursive code is basically the same

I might go even deeper with the caching but for this release this is pretty cool

@patdunlavey @aksm @giancarlobi (bug finder!) if any of you could please give this a try? This includes also the code from the open #189 but since it has not been merged it will show as "added" here. Once its merged (@patdunlavey could you test it already?) the actual file changes here will be way shorter

Please clear your caches.

Other Improvements
- Adds limits for how many Files will trigger reduced TechMD (so you end with smaller JSON)
- Gives asStructure Generation an extra argument (with a default so we do not need to change code somewhere else) in case we need to force reduced TechMD generation (when we know we have a lot of files but are not aware of how many)

Thanks


DiegoPino:
Ok, this is going in. No more changes to Strawberryfield for now. (except bug fixes)

PR #206, opened 2021-11-04
Output the arguments of ingesting the node as well as the file ingest…

ksuquix:
I saw that the arguments were being shown on the files getting ingested, but not the node itself.
This information is helpful not in just debugging, but understanding the process of ingestion if someone wants to create a pipeline to drop objects into the system externally with curl.


DiegoPino:
@ksuquix thanks =) (love this interaction!)
I changed the base of your commit to 1.0.0-RC3 which is where we go now. Can explain our release process better, but main is basically a stable snapshot of the last released version and once merged there nothing gets into it.

Good suggestion!

DiegoPino:
@ksuquix merged. Thanks for your contribution!!

Issue #207, opened 2021-11-05
Open range edtf dates cause errors when indexing


patdunlavey:
The problem:

Set a date with the value "../1800" (open range date with no start date) then try to index it as a date. You'll get an error:
```
Error: Call to a member function getMin() on null in EDTF\Model\Interval->getMin() (line 58 of /var/www/html/vendor/professional-wiki/edtf/src/Model/Interval.php).
```
This is because when the string is parsed, the returned object is an interval date with no start date. getMin on an interval date tries to getMin on the start date, which is null, throwing the error.

A solution is to check the class of edtf date that is returned by parsing, and if an interval date, first check to see if hasStartDate() (and hasEndDate()) before running getMin.

A PR to fix this will follow!

DiegoPino:
I see.. mmm..good catch! 🥼  I mean... how much EDTF do you all use? every possible variation?? Gosh. Will wait for the pull and this needs to go into the release so please hurry! 😆

patdunlavey:
PR created: https://github.com/esmero/strawberryfield/pull/208

DiegoPino:
@patdunlavey so this is a bug in EDTF?

Because the constructor for the Interval type does not allow a NULL start nor end? (of type IntervalSide) ?

```PHP
class Interval implements EdtfValue {
	use CoversTrait;

	public const NORMAL = 0;
	public const OPEN = 1;
	public const UNKNOWN = 2;

	private IntervalSide $start;
	private IntervalSide $end;
	private ?int $significantDigit;
	private ?int $estimated;

	/**
	 * @throws InvalidArgumentException
	 */
	public function __construct(
		IntervalSide $start,
		IntervalSide $end,
		?int $significantDigit = null,
		?int $estimated = null
	) {
		$this->start = $start;
		$this->end = $end;
		$this->significantDigit = $significantDigit;
		$this->estimated = $estimated;

		$this->assertAtLeastOneSideIsNormal();
		$this->assertEndIsLaterThanStart();
	}

	private function assertAtLeastOneSideIsNormal(): void {
		if ( !$this->start->isNormalInterval() && !$this->end->isNormalInterval() ) {
			throw new InvalidArgumentException( 'Interval needs to have one normal side' );
		}
	}

	private function assertEndIsLaterThanStart(): void {
		if ( $this->hasStartDate() && $this->hasEndDate() ) {
			// TODO: refactor to earlierThan/laterThan methods on ExtDate
[...]

PR #208, opened 2021-11-05
ISSUE-207 --- proposed fix for EDTF parsing problem with open range interval dates

patdunlavey:
also includes fix for cases when the jmespath_result is not a string (when the value in the field was json_decoded to an array).

See discussion here: https://github.com/esmero/strawberryfield/issues/207

### DIFF
--- a/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
+++ b/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
@@ -101,7 +101,7 @@ class StrawberryValuesViaJmesPathFromJson extends ItemList {
             $edtf_value = $result->getEdtfValue();
             switch(get_class($edtf_value)) {
               case "EDTF\Model\Interval":
-                if($result->getEdtfValue()->hasStartDate()) {
+                if($edtf_value->hasStartDate()) {
                   $values_parsed[] = date('c', $edtf_value->getMin());
                 }
                 if($result->getEdtfValue()->hasEndDate()) {
--- a/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
+++ b/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
@@ -85,6 +85,11 @@ class StrawberryValuesViaJmesPathFromJson extends ItemList {
       }
       // This is an array, don't double nest to make the normalizer happy.
       $jmespath_result_to_expose = array_filter($jmespath_result_to_expose);
+      foreach($jmespath_result_to_expose as $i => $v) {
+        if(!is_string($v)) {
+          $jmespath_result_to_expose[$i] = json_encode($v);
+        }
+      }
       $values = array_map('trim', $jmespath_result_to_expose);
       $values = array_map('stripslashes', $values);
       if ($is_date) {
@@ -93,12 +98,21 @@ class StrawberryValuesViaJmesPathFromJson extends ItemList {
         foreach ($values as $value) {
           $result = $parser->parse($value);
           if ($result->isValid()) {
-            $date_ini = $result->getEdtfValue()->getMin();
-            $date_max = $result->getEdtfValue()->getMax();
-            $date_ini = date('c', $date_ini);
-            $date_max = date('c', $date_max);
-            $values_parsed[] = $date_ini;
-            $values_parsed[] = $date_max;
+            $edtf_value = $result->getEdtfValue();
+            switch(get_class($edtf_value)) {
+              case "EDTF\Model\Interval":
+                if($result->getEdtfValue()->hasStartDate()) {
+                  $values_parsed[] = date('c', $edtf_value->getMin());
+                }
+                if($result->getEdtfValue()->hasEndDate()) {
+                  $values_parsed[] = date('c', $edtf_value->getMax());
+                }
+                break;
+              default:
+                $values_parsed[] = date('c', $edtf_value->getMin());
+                $values_parsed[] = date('c', $edtf_value->getMax());
+                break;
+            }
           }
           else {
             // If not EDTF (e.g an already ISO8601 date)
@@ -234,4 +248,3 @@ class StrawberryValuesViaJmesPathFromJson extends ItemList {
     return FALSE;
   }
 }
-
--- a/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
+++ b/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
@@ -86,7 +86,7 @@ class StrawberryValuesViaJmesPathFromJson extends ItemList {
       // This is an array, don't double nest to make the normalizer happy.
       $jmespath_result_to_expose = array_filter($jmespath_result_to_expose);
       foreach($jmespath_result_to_expose as $i => $v) {
-        if(!is_string($v)) {
+        if(is_array($v) or is_object($v)) {
           $jmespath_result_to_expose[$i] = json_encode($v);
         }
       }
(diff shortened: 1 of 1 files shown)


src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
```diff
@@ -85,6 +85,11 @@ public function process($langcode = NULL)
       }
       // This is an array, don't double nest to make the normalizer happy.
       $jmespath_result_to_expose = array_filter($jmespath_result_to_expose);
+      foreach($jmespath_result_to_expose as $i => $v) {
```
DiegoPino, line 88: Wait. What is this?

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
```diff
@@ -93,12 +98,21 @@ public function process($langcode = NULL)
...
           if ($result->isValid()) {
-            $date_ini = $result->getEdtfValue()->getMin();
-            $date_max = $result->getEdtfValue()->getMax();
-            $date_ini = date('c', $date_ini);
-            $date_max = date('c', $date_max);
-            $values_parsed[] = $date_ini;
-            $values_parsed[] = $date_max;
+            $edtf_value = $result->getEdtfValue();
+            switch(get_class($edtf_value)) {
+              case "EDTF\Model\Interval":
```
DiegoPino, line 103: Ok, I do not get this part. And if it is not an interval?
you still do
$values_parsed[] = date('c', $edtf_value->getMin());
$values_parsed[] = date('c', $edtf_value->getMax());
?

A bit confused, could you explain?

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
patdunlavey: that's "also includes fix for cases when the jmespath_result is not a string" - unrelated to this issue. Reject if you prefer.

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
patdunlavey: This could just be an if/else, but when I wrote it I wasn't sure what other classes that $result->getEdtfValue() might return, so I did it as a switch. Default is to do it as before, but if the resulting edtf date is an interval, then check for presence of start and end dates before trying to getMin/Max.

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
DiegoPino: No, not rejecting but trying to figure out how this would work? So you mean if not a string, you want to actually index the JSON string?

DiegoPino: Also, if new please add a comment on top (before the foreach) explaining the expected behavior.

This scares me a bit because if you try to index let's say as:image... it would for each image the full JSON as a multivalued field? Huge string?

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
DiegoPino: Ok. That makes sense, So again. This is a bug in `EDTF EDTF\Model\Interval` because as explained in the issue the constructor of that class does not allow an empty/null start nor end... so it should always have a return... interface also applies to every type of edtfValue. Did I get that right?

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
```diff
@@ -93,12 +98,21 @@ public function process($langcode = NULL)
...
+                if($result->getEdtfValue()->hasStartDate()) {
```
DiegoPino, line 104:
```suggestion
                if($edtf_value->hasStartDate()) {
```

DiegoPino, line 104:
```suggestion
                if($edtf_value->hasStartDate()) {
```

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
DiegoPino, line 103: Could you add a small //@TODO remove once EDTF fixes their invalid Constructor for EDTF\Model\Interval that should per interface never allow NULL for start nor end date

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
```diff
@@ -85,6 +85,11 @@ public function process($langcode = NULL)
       }
       // This is an array, don't double nest to make the normalizer happy.
       $jmespath_result_to_expose = array_filter($jmespath_result_to_expose);
+      foreach($jmespath_result_to_expose as $i => $v) {
+        if(!is_string($v)) {
```
DiegoPino, line 89: May be also a number friend... Do we want to check for (is_array(..) || is_object(..)) instead?

DiegoPino (changes requested):
Ok, I get this now. I asked for a few changes. I need to be convinced about JSON encoding values as strings in case the JMESPATH is a nested thing ... this may lead to "undesired" indexing...

DiegoPino:
@patdunlavey I get all now. Please convince me of the need of Indexing the JSON in case of a deeply nested Array is the result of the JMESPATH. I can understand the use case? But also it may lead to unexpected large Solr Documents, so I need some use cases needs here. I added a few comments and a code suggestion. Thanks a lot!!!!

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
DiegoPino: I think it may only be array since we json_decoded but just to be sure ...

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
patdunlavey: yes I think you get it

patdunlavey:
I tried to address each of your suggestions/requests. Can you re-review @DiegoPino ?

patdunlavey:
Oh, regarding "Please convince me of the need of Indexing the JSON in case of a deeply nested Array is the result of the JMESPATH. I can understand the use case? But also it may lead to unexpected large Solr Documents, so I need some use cases/needs/rationales here". At [this line](https://github.com/esmero/strawberryfield/blob/main/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php#L87) we assume the values are strings. So the use-case is simply that we need to ensure that the values are strings. I was seeing errors here, which is what prompted the suggested change.

DiegoPino:
@patdunlavey o weird. So this
https://github.com/esmero/strawberryfield/blob/main/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php#L75
Is not working then? Ok, can you remove that piece (for now or I can merge over once this is merged, just give me a sign), I will get the "is multi item" code fixed instead because that should fix it at the base method. I appreciate your work on this. All rest looks perfect and will merge ASAP.

src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php
DiegoPino: Let's roll with this, I should really fix StrawberryfieldJsonHelper::arrayIsMultiSimple instead (may in a next commit or we can roll with your fix until 1.1.0

patdunlavey:
I'm not sure, but I guess it must not be working, if its purpose is to ensure that when we get to https://github.com/esmero/strawberryfield/blob/dfb534eaa8e07f76677b1fa56337fdb07d1c7a66/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php#L87 it will be an array of strings. I wasn't sure what it was intended to do there.

DiegoPino (approved):

patdunlavey:
Thank you @DiegoPino !!

Issue #222, opened 2022-05-18
Add a Garbage Collection Service driven by event


DiegoPino:
# What?

Temporary files can grow huge. But our primary motivation is to keep ingest safe and your assets in the right spot. 

This work will implement a safe method to deal with left overs (composting) via an Event Subscriber that will handle Garbage/temp files on any operation that generates left overs. 
- The Event Subscriber will enqueue those files metadata (creation date) and module ingesting those. 
- A queue worker will run via Hydroponics checking against config if the files in the queue are still valid (time based) and/or used by a File Entity. If they qualify for removal they will be totally removed from Filesystem. If not re-enqueued for later processing.

- AMI will trigger the event on "File Download"
- Metadata Service will trigger the event on "Ensure Local"
- Strawberryfield will trigger the event on "File Persistence" (when copied to e.g S3)
- Strawberry Runners will trigger the event on "Temp file generation for OCR", etc.

If a file in the queue is not found it will be removed silently from the queue.

We might also allow Modules to define their persistence (temp folder) location to allow a "check all existing temp files" in case the queue as manually emptied.

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/7e07661ef57ad228d36b08c312fed1d6a1c2ee26

PR #223, opened 2022-05-19
ISSUE-222: Composer bin service. Enqueue and delete temp files

DiegoPino:
See #222 

This pull adds:

- New EventType and Event Subscriber. Event is triggered during and ADO's file persistence and when we are sure it happened. The Subscriber fills a queue.
- The queue worker checks if the file is there, if its safe, if its not used by an Entity, if it has been longer or === than the max time to live and deletes it. If all matches, except time it re-enqueues (Double check Hydroponics since we are using a new D9 Class here)
- There is also a hook that allows other modules to inform where/what is safe.
- There is also a new Config to be seen at `/admin/config/archipelago/filepersisting`
image

The queue needs to be configured to Run hopefully using Hydroponics but testing can be also done via the System Queue manager.

Other modules (e.g SB Flavors) will invoke the event too to mark its own temps are "compostable"♻️ 

@patdunlavey @alliomeria @aksm @giancarlobi If any of you see anything strange in the code let me know?

### DIFF
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -258,6 +258,6 @@ strawberryfield.breadcrumb_settings:
       label: 'Type of Breadcrumb to process for ADOs'
       default: 'longest'
     enabled:
-      type: boolan
+      type: boolean
       label: 'If Breadcrumb processing is enabled or not'
       default: TRUE
--- a/config/schema/strawberryfield.schema.yml
+++ b/config/schema/strawberryfield.schema.yml
@@ -93,6 +93,10 @@ strawberryfield.storage_settings:
       type: string
       label: 'Storage Path under the Storage Scheme, for Persisting Digital Objects in JSON format'
       default: "dostorage"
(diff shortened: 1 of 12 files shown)


patdunlavey:
Hi @DiegoPino , I'm just catching up on your amazing work. I absolutely :heart: this! Thank you so much! Initial testing seems to be good.

The only fly in the ointment that I see is that for temporary AMI files it leaves the empty directories. I'm not sure if there's an easy solution for that. After the [$this->filesystem->delete($uri)](https://github.com/esmero/strawberryfield/pull/223/files#diff-faededbc7fdb07322fd7d820ec72d91982162ab8f1d5d5a9265792d0a686b7b9R270), might it make sense to check to see if there are any remaining files in the deleted file's parent directory, and if not, remove the directory?

DiegoPino:
Interesting! Good catch. Yes, that is a good idea. Can you open an issue? I can code that. Thanks so much. Also, maybe we should add an extra option that is lower than 6 hours?

PR #229, opened 2022-08-04
ISSUE-228: node_type deprecated and removed in Drupal 10

### LINKED ISSUE #228: node_type deprecated and removed in Drupal 10
# What?

`node_type` is deprecated and removed in Drupal 10. The pathauto module has an update hook that takes care of this, but it's possible (happened to me!) to run into the weird situation where a fresh install of the newer pathauto versions runs the update hook, which is then negated by an import of the older configs. We can offset this situation by providing the same update hook, which will run when they update our module.

Issue for reference: https://www.drupal.org/project/pathauto/issues/3222776

aksm:
Starts to resolve #228 

Additional commits to come.

### DIFF
--- a/strawberryfield.install
+++ b/strawberryfield.install
@@ -55,3 +55,26 @@ function strawberryfield_update_9101() {
     }
   }
 }
+/**
+ * Update node type conditions from node_type to entity_bundle.
+ * (Copied from pathauto.install)
+ */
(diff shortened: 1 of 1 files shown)


DiegoPino:
Will merge and work from here to make the block part editing generic. Thanks @aksm

PR #235, opened 2022-10-21
ISSUE-234: Make Flavor search aware of CWS/Children based OCR

DiegoPino:
Still WIP don't even test

### DIFF
--- a/src/StrawberryfieldUtilityService.php
+++ b/src/StrawberryfieldUtilityService.php
@@ -336,17 +336,32 @@ class StrawberryfieldUtilityService implements StrawberryfieldUtilityServiceInte
         'limit' => 1,
         'offset' => 0,
       ]);
+
+      $allfields_translated_to_solr = $index->getServerInstance()
+        ->getBackend()
+        ->getSolrFieldNames($index);
       $parse_mode = $this->parseModeManager->createInstance('terms');
       $query->setParseMode($parse_mode);
       $query->sort('search_api_relevance', 'DESC');
 
       /* Forcing here two fixed options */
       $parent_conditions = $query->createConditionGroup('OR');
-      $parent_conditions->addCondition('parent_id', $entity->id());
-      $parent_conditions->addCondition('top_parent_id', $entity->id());
 
-      $query->addConditionGroup($parent_conditions)
-        ->addCondition('processor_id', $processor)
+      if (isset($allfields_translated_to_solr['parent_id'])) {
+        $parent_conditions->addCondition('parent_id',  $entity->id());
+      }
+      // The property path for this is: target_id:field_descriptive_metadata:sbf_entity_reference_ispartof:nid
+      // TODO: This needs a config form. For now let's document. Even if not present
+      // It will not fail.
+      if (isset($allfields_translated_to_solr['top_parent_id'])) {
+        $parent_conditions->addCondition('top_parent_id',  $entity->id());
+      }
+
+      if (count($parent_conditions->getConditions())) {
+        $query->addConditionGroup($parent_conditions);
+      }
+
+      $query->addCondition('processor_id', $processor)
         ->addCondition('search_api_datasource',
           'strawberryfield_flavor_datasource');
 
--- a/src/StrawberryfieldUtilityService.php
+++ b/src/StrawberryfieldUtilityService.php
@@ -340,7 +340,12 @@ class StrawberryfieldUtilityService implements StrawberryfieldUtilityServiceInte
       $query->setParseMode($parse_mode);
       $query->sort('search_api_relevance', 'DESC');
 
-      $query->addCondition('parent_id', $entity->id())
+      /* Forcing here two fixed options */
+      $parent_conditions = $query->createConditionGroup('OR');
+      $parent_conditions->addCondition('parent_id', $entity->id());
+      $parent_conditions->addCondition('top_parent_id', $entity->id());
+
+      $query->addConditionGroup($parent_conditions)
         ->addCondition('processor_id', $processor)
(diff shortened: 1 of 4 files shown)


DiegoPino:
@patdunlavey this is the code that solves an 85% of the use cases. Will add a guide with screenshots tomorrow but the code as it is should be ready. The other part goes into `format_strawberryfield` since metadata displays are defined there and I can not introduce a bidirectional dependency between modules.

src/StrawberryfieldUtilityService.php
```diff
@@ -340,7 +340,12 @@ public function getCountByProcessorInSolr(
       $query->setParseMode($parse_mode);
       $query->sort('search_api_relevance', 'DESC');
 
-      $query->addCondition('parent_id', $entity->id())
+      /* Forcing here two fixed options */
+      $parent_conditions = $query->createConditionGroup('OR');
```
DiegoPino, line 344: This here needs and extra check ... more tomorrow!

DiegoPino:
@patdunlavey,@alliomeria and @karomabiles (if you want to see this working) instructions for testing OCR for compounds:

- Get this branch (docker exec -ti esmero-php bash -c "composer require 'strawberryfield/strawberryfield:dev-issue-234 as dev-1.0.0'"
- Clear caches.
- Add the following fields to your Strawberry Runners Data Sources at `/admin/config/search/search-api/index/default_solr_index/fields`
image
- Important is that both "Machine names" and "property paths are those", if in doubt let me know
- Make sure that your Strawberry Runner HOCR (the pager and the OCR) are processing image/tiff, image/jpeg, application/pdf and targeting also "painting" (all this is for the example)
- Generate a Top New ADO of type collection (a Creative Work Series one) with basic metadata.
- Add 3 Digital Object ADOs each one with e.g a single JPEG using "Painting as type" and make them Children of the First (CWS one)
  1. [page1](https://user-images.githubusercontent.com/6946023/199346284-60b8a9ee-5a53-49f3-bedf-901582d6fd9c.jpg)
  2. [page2](https://user-images.githubusercontent.com/6946023/199346292-ae2ebba8-0f56-4125-88b4-3c1a7b8e58a8.jpg)
  3. [page3](https://user-images.githubusercontent.com/6946023/199346295-a5f110f3-6f1f-4fe9-bf83-d3d3dba4edc6.jpg)

And make sure each one also has "sequence_id" set, to 1, 2 and 3 respectively. If your webforms don't have that element/key (we need it) please add it or edit the JSON RAW. Save.

Make sure the Queue is processed (All Background ones that will generate OCR).

- Now make sure you have a special View Mode for testing. 

Mine is named:

image

and has these settings:

image

Basically you want to have the IABookreader but using the IIIF V3 CWS as template as source. Now Apply that view Mode to the Top Object by editing and forcing that Display Mode
image

You should not need to reindex at this stage (if you followed this steps for this demo object)

Search for "Queen", "Pumpkin" and  "King". Each should be highlighted correctly on its own page. Now search for "OCR" multiple pages.

This covers the basic use case where all children have a `sequence_id` and all are shown in the Manifest.
Still working on the complex (a setting) use case where the structure shown is different, maybe only `odd` pages, etc.

Please let me know if you have issues/questions/needs

patdunlavey:
@DiegoPino starting to look at this now (sorry for the delay!!!)

patdunlavey:
@DiegoPino I was able to reproduce your steps, and your result! The only problem I noticed is that I don't get the pins in the result bar. I suspect that's due to me not being fully caught up to changes in the IIIF Presentation API 3 Creative Works Series Manifest.

I tested what happens when I add a second image file to one of the child objects. It seems to OCR correctly, but it is saved in the key_value table with the sequence number of "1", rather than that found at `"as:image".*.sequence`. As a result, when I display in the bookviewer, I get the additional page, but highlighting is off. In this case, I added your sample image file to a page in this object, and though it searches successfully for it (the word "queen" in this example), it highlights on the wrong page: 
image

Not sure if this is a simple problem to solve (and whether it's in the 15% you referred to!).

patdunlavey:
Looking [here](https://github.com/esmero/strawberry_runners/blob/main/src/Plugin/StrawberryRunnersPostProcessor/JsonFileSequencePostProcessor.php#L179), it seems like the sequence number should be correct. Not sure why it isn't!

DiegoPino:
@patdunlavey adding a new page and having key_value = 1 is OK. I wonder if you added the "sequence_id" JSON KEY key to your new page/ADO?

DiegoPino:
The actual page matching here depends on having a sequence_id at at Child ADO level. Without it, the Manifest is going to show pages in any order and won't match the response (and re-lative new ordering of results from the search) order that happens here now. The re-paging of the results happens here:
https://github.com/esmero/strawberryfield/blob/3d022aeb07a85bd39c477790669ee1254f275fc2/src/Controller/StrawberryfieldFlavorDatasourceSearchController.php#L290
So if your ADO (the one that produced the HOCR) has no sequence_id it will return 1 and thus will offset all. Your new page should have sequence_id = 4 (in the JSON) now

DiegoPino:
Also, the lack of pins in the result bar is strange. Are you using this on top of a custom code piece? e.g have you started modifying any other part of Archipelago already? Weird because on a fresh 1.0.0 I do see the pins .... maybe we need to have a call!

DiegoPino:
@patdunlavey will merge and we open a new Pull/ISSUE for troubleshooting? There is more work to be done on SBFlavors for sure and I can add any corrections to a new pull.

patdunlavey:
Sorry, I meant to get the results of my investigation in earlier! I'll make a new ticket for the multi-file sequencing issue.

PR #238, opened 2022-11-04
ISSUE-237: Move ADO Tools into this module, make JMESPATH a subsection + add a view for listing children

### LINKED ISSUE #237: Move ADO Tools into this module, make JMESPATH a subsection + add a view for listing children
# What?

ADO TOOLS tab (form/controller) is not on the right module and JMESPATH should be ONE of many tools available for admins.

WE need
- Move the code/classes/permissions/routes to this module from strawberry_runners
- Make JMESPATH tester a subsection (a tab) on the main ADO Tools route (using one of the other tasks/menu/action YML files)
- Add a Modal "Children ADOs" Tab that uses a View (and thus has a controller) that lists all Children (or related objects so can be more tan a single property... might be ispartof initially but since it will be a view might be more) of the Current NODE with a similar layout that /admin/content (the list of nodes). This View should be shipped with our module and be very simple. We can then extend/add/replace with more advanced ones. 

Eventually the later view could be adapted/modified/rethemed to allow dragging/for sorting or do other fancy stuff. For now a listing is already a lot/good

This is @aksm holiday assignment!

aksm:
Starts to resolve #237.

Move existing ADO Tools menu and JMESPATH codemirror library and update any references. This just moves everything over from strawberry_runners while retaining the exact same functionality (tested).

esmero/strawberry_runners#63 should be merged before merging this.


aksm:
@DiegoPino I had merged in recent changes up to that point. Should I merge in the further changes since?

> @aksm hi. This is bringing commits that don't belong here. Are those OK? e.g [src/Controller/StrawberryfieldFlavorDatasourceSearchController.php](https://github.com/esmero/strawberryfield/pull/238#diff-f1cd623d81528b2c011f3f8fc44d272952fd20bb0d07eabe70cdae374419ab57)
> 
> Can you please double check/rebase and cleanup? Thanks

DiegoPino:
@aksm normally you don't need to merge other changes. You fetch a branch, create a new one from that one and you only modify what you need too. Only reason to merge other changes is when you are touching those files. Here you can rebase to 1.1.0. Or are these changes in 1.0.0 not present in 1.1.0?

src/Ajax/UpdateCodeMirrorCommand.php
```diff
@@ -0,0 +1,51 @@
...
+ * Created by PhpStorm.
+ * User: dpino
+ * Date: 10/1/19
+ * Time: 11:30 AM
+ */
+
+namespace Drupal\strawberryfield\Ajax;
+use Drupal\Core\Ajax\CommandInterface;
+class UpdateCodeMirrorCommand implements CommandInterface
+{
```
DiegoPino, line 12: Move bracket to the same line as the class

src/Form/StrawberryfieldToolsForm.php
```diff
@@ -0,0 +1,157 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Ajax\AjaxResponse;
+use Drupal\strawberryfield\Ajax\UpdateCodeMirrorCommand;
+
+/**
+ * Returns responses for Node routes.
+ */
+class StrawberryfieldToolsForm extends FormBase {
+
+  /**
```
DiegoPino, line 20: Line 20 to line 57 can one removed. This is form is only using what FormBase is already providing. Please test (clearing caches bc Forms are cached)

src/Form/StrawberryfieldToolsForm.php
```diff
@@ -0,0 +1,157 @@
...
+              '#executes_submit_callback' => TRUE,
+              '#submit' =>  ['::submitForm']
+            ];
+            $form['test_jmespath_input'] = [
+              '#type' => 'codemirror',
+              '#codemirror' => $settings,
+              '#default_value' => $json,
+              '#rows' => 15,
+            ];
+            $form['test_output'] = [
```
DiegoPino, line 102: Please move the output to before the actual Input

DiegoPino (changes requested):
Small changes. Sorry, those were wrong from the old code!

Thanks for your patience (excellent service!)

src/Ajax/UpdateCodeMirrorCommand.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: dpino
+ * Date: 10/1/19
+ * Time: 11:30 AM
+ */
+
+namespace Drupal\strawberryfield\Ajax;
+use Drupal\Core\Ajax\CommandInterface;
```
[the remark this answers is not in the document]
aksm: Resolved by e5c480833d0a9dde6f6e6385b60a4a363516ca1c

src/Ajax/UpdateCodeMirrorCommand.php
aksm: Resolved by e5c480833d0a9dde6f6e6385b60a4a363516ca1c

src/Form/StrawberryfieldToolsForm.php
aksm: Resolved by 99a8e21bf3275f9956960135b93aef4c5df5d120

src/Form/StrawberryfieldToolsForm.php
aksm: Resolved by 50b5e743edb9cc7639263086697a9586340a4aef

src/Form/StrawberryfieldToolsForm.php
DiegoPino: @aksm hi, this is not resolved yet. The constructor is still using  it (and still present, we don't need it)

again, we can delete __construct and create here since there is nothing we need to override from the base form

DiegoPino (changes requested):
Still a few unresolved issues

src/Form/StrawberryfieldToolsForm.php
aksm: @DiegoPino Sorry, thought I had already removed those but missed them. Now resolved in 652bd8e774941c04c854c869a8ebaf33b644f67e

aksm:
@DiegoPino Will make a separate PR for the views needed for this, but here is what I used for developing/testing:

[ado_tools_children_views.zip](https://github.com/esmero/strawberryfield/files/11020486/ado_tools_children_views.zip)

aksm:
Hi @DiegoPino, except for the views (see above comment), I think this is ready for review.

DiegoPino:
@aksm thanks. Will review during the day!

DiegoPino:
@aksm regarding the default views.
I would explore using the config/install folder of this module and see also this function `search_api_solr_update_helper_install_configs` (as a reference) which should allow us to deploy selectively the new views via an update hook (v/s having to uninstall/reinstall the module)

aksm:
@DiegoPino I basically copied the function you referenced above as is in order to push the commits in by EOD Friday (which failed anyway because I was in a detached head state and ended up coming back later in the evening here: 00ec5c9a3f0c1a03c27de2d79ce28cb68d5a9224), and it seemed to work when I tested, but I'm glad to keep on it if you think there's more needed or if anything's missing. Either way, I'm sure the views will need more tweaking.

Issue #247, opened 2022-12-12
Bug on SBF datasource parent iterator tracker + legacy RC1 OCR fix


DiegoPino:
See #244

For some reason (December, me distracted) I introduced bug where instead of assigning a value to a new Array Key to track dependent NODE updates, I ended fetching one that is not yet there. The difference is subtle and I blame the stupid max 80 characters line DCS and me getting old

This also adds a workaround to remove empty OCR tags coming from RC1 (or generated in RC1). This affects OCR done with older archipelagos that the most recent (but not the latest) version of the OCR Plugin can not manage. More work on this will be made but for now its good enough

DiegoPino:
Fixed via https://github.com/esmero/strawberryfield/commit/9792aed7f2f3d3e2128a1baac3d0f36bcff606af

Issue #244, opened 2022-11-21
Trigger a parent Node (or parent parent) Index tracker update on last sequence of a SBF + Computed field


DiegoPino:
# What?

We need to aggregate Strawberry Flavors at the ADO level for unified search to happen (performance reasons means a subquery is a bad idea given how Drupal Views work related to Search API functionality...)

The idea is basic. An ADO Solr Document is ready way before all their Flavor Children are ready. So we need a performant way (means not on every Child generation) to tell Solr that the parent needs to be reindexed so it can gather all the little children at once.

This also means we need a Computed (global field?) That does the gathering:
Two options:
- Use the same idea of the one that already exist and renders a View Mode, but this one will do "a Solr query" to gather all Text only Values from all Flavors.
- Add logic to the Key Name provider we already have for this and do the logic there. Means expose the Text as a Field property.

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/3a8f0374567ba38b1ae0fd52ed241b0de962a2a6

PR #248, opened 2022-12-12
ISSUE-247: Fix bug on parent of parent and cleanup empties for legacy OCR

DiegoPino:
See #247 and #244 

- The legacy OCR cleanup is a temp fix for WIP 1.1.0. Soon this will also be commited back into the backend AND will be only a workaround.

- The bug is(I am) stupid, instead of assigning the `$referencedEntity` to a new key i was retrieving a non existing key! December...


DiegoPino:
@alliomeria this needs to be merged and applied to a bunch of repos. Won't affect anything (the bug) as it is now but will throw a few warnings. The OCR fix allows old OCR (e.g studio) to be indexed correctly into 1.0.0's newer Solr plugin

aksm:
@DiegoPino This looks good to me!

PR #252, opened 2022-12-16
ISSUE-251: Simulate message in StrawberryfieldJsonPatch in wrong place

### LINKED ISSUE #251: Simulate message in StrawberryfieldJsonPatch in wrong place
# What?
In the below line, the message gets triggered whether simulating or not and has the simulate configuration value, which is an int.

https://github.com/esmero/strawberryfield/blob/7166e579434dc0590a5fab2c0b1f2bcf2a9719c8/src/Plugin/Action/StrawberryfieldJsonPatch.php#L218

The message needs to be moved to the simulate check further down and should match the one for text-based, i.e.:

https://github.com/esmero/ami/blob/f6bf9dd33ad94a44255fe2e30c031ef9f50a0b31/src/Plugin/Action/AmiStrawberryfieldJsonAsText.php#L137-L138

aksm:
Resolves #251. Move simulate message to if simulate block and update message to match text-based action in AMI for consistency.

### DIFF
--- a/src/Plugin/Action/StrawberryfieldJsonPatch.php
+++ b/src/Plugin/Action/StrawberryfieldJsonPatch.php
@@ -215,7 +215,6 @@ JSON;
       if ($sbf_fields = $this->strawberryfieldUtility->bearsStrawberryfield(
         $entity
       )) {
-        $this->messenger()->addMessage($this->configuration['simulate']);
         $this->patchArray = json_decode($this->configuration['jsonpatch']);
         foreach ($sbf_fields as $field_name) {
           /* @var $field \Drupal\Core\Field\FieldItemInterface */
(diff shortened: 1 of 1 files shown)


src/Plugin/Action/StrawberryfieldJsonPatch.php
```diff
@@ -242,6 +241,7 @@ public function execute($entity = NULL) {
                 );
               };
               if ($this->configuration['simulate']) {
+                $this->messenger()->addMessage('In simulation Mode');
```
DiegoPino, line 244: Hey, thinking about this twice (sorry, slow), better just remove this. Because on 1000 objects this is basically 1000 messages right? Or maybe it can be send before the execute is send? like on page 1999? What do you think?

aksm: Ah ok, that makes sense. Will just remove for now.

DiegoPino (approved):

DiegoPino (approved):

Issue #257, opened 2023-02-18
Search API Excerpts use wrong Cache tags and thus only deliver when called via AJAX/POST


DiegoPino:
# What?

Excerpts that are shown on Views results and Rendered ADOs (or nodes) use only the URL context for caching, and because of that when not called via POST/AJAX (which bypasses caching) get stuck on basically not displaying anything at all when e.g a URL with the search term is called directly (the view Cache wins there). 

# How to get around this?

I see two options here:
- Use a promise of a render (basically a dynamic rendered or also named in Drupal #lazy_builder)
- Change in case Excerpts are present the global cache of a view (always adding the URL initially as a content + Max Cache time: 0 when the excerpt is produced.

@alliomeria this relates to one of our partners questions.

DiegoPino:
See:

```PHP
/**
 * Implements hook_entity_view().
 */
function search_api_entity_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
  $excerpt_component = $display->getComponent('search_api_excerpt');
  if ($excerpt_component !== NULL && isset($build['#search_api_excerpt'])) {
    $build['search_api_excerpt'] = [
      '#type' => 'markup',
      '#markup' => $build['#search_api_excerpt'],
      '#cache' => [
        'contexts' => ['url.query_args'],
      ],
    ];
  }
```

Another option would be (without code) to test if instead of rendering an Entity on the results we can use the "rendered entity" field and output the excerpt as an additional field. That way the excerpt is generated by the Row plugin in Views instead of the actual hook_entity_view() for nodes/view mode

DiegoPino:
See also https://www.drupal.org/docs/drupal-apis/render-api/cacheability-of-render-arrays

PR #258, opened 2023-02-21
ISSUE-257: Humble (and experimental) attempt at dealing with stale/missing excerpts

DiegoPino:
See #257 

... on GET calls.
This enables a new config option for our advanced highlighter and enables a very experimental lazy loading mechanism when needed, allowing still POST calls to get around the fact that they are "barely" cached... (except at the Views level, but we don't mess with that right? :( Also this whole day could have been a cache time set to 0 .... but i like people...

To be honest, I feel this is over engineered... I understand the use case and the need and I might develop this better/further soon. I don't know if even caching this beyond time to live 0 is more performant or not


DiegoPino:
This requires https://github.com/esmero/format_strawberryfield/issues/287 to see it in action, but won't fail if not merged

Issue #265, opened 2023-04-11
Composer won't delete base file names I considered unsafe (weird issue)


DiegoPino:
# What?

So this is not really an issue. I mark files that start with a `.` and end with `.php` etc as unsafe to be composted. We have seen an Archipelago not composting lately, happens that all files being remotely downloaded are of this type `..myfile.pdf`

I would dispute the file naming convention here (a lot) but the fact is we need to compost these.
So, I need ideas/solutions.

Lower the security? Do it differently?

The code is here:

https://github.com/esmero/strawberryfield/blob/573ffa44a369ad68c59a92b2746258c2671ef13f/src/Plugin/QueueWorker/CompostQueueWorker.php#L218-L236

@alliomeria and @karomabiles  this affects you directly
@aksm ideas? (other than removing the [.] as unsafe)

aksm:
@DiegoPino Is it possible to do an override for file name patterns that are ok to delete?

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/041183e535d985824721aba2d87a4d896030fc0c

PR #266, opened 2023-04-11
ISSUE-265: Allow a workaround for dot files when composting

DiegoPino:
see #265 

This will also always re-enqueue unsafe files and log them. This will at least allow us to show the user these exist and allow an informed decision to be made. @alliomeria @aksm those this sound ok?

Could I get a sanitizing look from you both?


aksm (approved):

DiegoPino:
@aksm @alliomeria did extensive testing. This is working well know. `dot` files are removed correctly with this setting enable, re-enqued and logged as "unsafe" if the setting is not present. Merging!

Issue #270, opened 2023-05-25
Allow also the shorter path to be used for Nodes / Breadcrumbs


DiegoPino:
# What?

Sometimes, when e.g an Object has multiple paths and there is commonalities (like e.g there is a common parent) the smart one might be too much (miss leading). So we add the option for the shorter one too

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/d498c09c828beebb34411720f2ed36b58162a507

PR #271, opened 2023-05-25
ISSUE-270: Enable the shortest (third option) Tarth for breadcrumbs

DiegoPino:
See #270 

Super simple. Not much more to say

Thanks @alliomeria for asking for this


alliomeria:
Thanks @DiegoPino for adding this! :)

Issue #297, opened 2024-02-06
Missed a Jquery Once() replacement in Key name Provider overview for Drupal 10


DiegoPino:
# What?

/admin/structure/strawberry_keynameprovider/overview uses Ajax/JS and I missed replacing a legacy jQuery Once with the new Drupal once() API. 

Fixing it. Will make a 1.3 and 1.4 fix

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/12dbc38ad7f5d7b1640aeb31805d0c5ef7a47533
and https://github.com/esmero/strawberryfield/commit/8a31b805363409e3dde4e7ec11b4dd0dd12d780e

Thanks for watching

PR #298, opened 2024-02-06
ISSUE-297: D3JS once v/s Jquery Once fix + small supporting code fixes

DiegoPino:
See #297 

This fixes it.Thanks @alliomeria

PR #299, opened 2024-02-06
ISSUE-297: For 1.4.0, fixes once on D3JS

DiegoPino:
See #297


DiegoPino:
Same as #298 , works so merging

Issue #300, opened 2024-02-07
Deprecation error: "trim" receiving NULL


patdunlavey:
Seeing a lot of these messages while indexing content:
```
Deprecated function: trim(): Passing null to parameter #1 ($string) of type string is deprecated in 
Drupal\strawberryfield\Plugin\DataType\StrawberryValuesFromJson->process() (line 76 of  
/var/www/html/web/modules/contrib/strawberryfield/src/Plugin/DataType/StrawberryValuesFromJson.php)
```
https://github.com/esmero/strawberryfield/blob/8a31b805363409e3dde4e7ec11b4dd0dd12d780e/src/Plugin/DataType/StrawberryValuesFromJson.php#L76

It presumes that every value in the array `$flattened[$needle]` is a string, where evidently that may not be the case. 

Is the fix for this to simply catch cases where the value is null and either remove it, or convert it to an empty string? Or is this indicative of a condition that should not occur?

DiegoPino:
Hi, thanks for reporting. Give me a few minutes to debug, we have plenty of archipelago's running this code base without erroring and since i was sure (no longer) i had fixed all trims i saw in November i want to be sure there is not another issue somewhere. Will report back in 30 min with a fix. Thanks!

DiegoPino:
@patdunlavey ok. It does not hurt to check for a null before trimming or even better, only trim strings. In my memory the flattener would had already dealt with empty values but seems like it is not. I will check if this is also a behavior that happens somewhere else. Pull coming in 15. Thanks

DiegoPino:
Fixed via https://github.com/esmero/strawberryfield/commit/3300fe6fd719b364a060d72808367a3dfe61f2ed and
https://github.com/esmero/strawberryfield/commit/8b47244c79f51e3a6b49694bd6073ac67a4bd095

PR #301, opened 2024-02-08
ISSUE-300: Safer trim on JSON values from key name provider

DiegoPino:
See #300 

@patdunlavey can you please test/see/look at the code and see if anything is amiss?

And if all works, approve? Thanks!

### DIFF
--- a/src/Plugin/DataType/StrawberryValuesFromJson.php
+++ b/src/Plugin/DataType/StrawberryValuesFromJson.php
@@ -71,12 +71,21 @@ class StrawberryValuesFromJson extends ItemList {
             // @TODO should we allow unicode directly?
             $item = json_encode($item,JSON_UNESCAPED_UNICODE);
           }
+          elseif (is_string($item)) {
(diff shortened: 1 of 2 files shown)


patdunlavey (commented):
Hi Diego, Thanks very much for your quick attention to this. I'll have to do a bit of setup to be able to test it, and that's not going to happen today. Just reading your code changes, while I don't understand the logic completely (what is the meaning of a FALSE value in this context, for example), it appears to solve the problem of trying to trim a NULL value.
I'll test properly as soon as I can.

DiegoPino:
@patdunlavey thanks. The extra checks on FALSE and 0 are because an array filter will remove actual 0 and booleans from the JSON and there might (they are really) needs of being able to index flags/boolean data into Solr.

I have a test suit for this. Will just ingest a few objects with empty arrays/nulls, etc and if all goes well merge. Thanks

patdunlavey:
Great! Thank you!!!

Issue #303, opened 2024-02-11
Escape tags on the default raw json formatter


DiegoPino:
@alliomeria maybe this is one you could tackle during the week? Same json argument we tested live with good results on Friday.Please ping me if you decide so(or if not too). For 1.4.0. Thanks!

alliomeria:
Hello @DiegoPino, following up on our live discussion from earlier today, since you are going to do the pull for code... I'll make a PR for the [corresponding template updates from my local deployment fork](https://github.com/esmero/archipelago-deployment/compare/1.3.0...alliomeria:archipelago-deployment:1.3.0) if that sounds ok to you. Let me know what you think. Thank you! :)

DiegoPino:
Thanks for the template update. Will close this ticket once you have that.

DiegoPino:
Solved via https://github.com/esmero/strawberryfield/commit/b6f56be08f16b0ecda0abc97e4bcbd1de6bf6372 and via template update by @alliomeria via https://github.com/esmero/archipelago-deployment/commit/4a381851debb9ebac938bbc34e2467747700d15d

PR #304, opened 2024-02-15
ISSUE-303: Escape Tags on RAW JSON formatter

DiegoPino:
# What?

See #303 

Helps fixing issues that kick you (bite and never let you go) when people add  to their JSON HTML tags inside values (and unclosed ones) and the RAW output ends breaking the JS (that is loaded HTML wise after the whole body ..)

Simple little one / smallish fix


alliomeria:
Looks great, thanks @DiegoPino !

DiegoPino:
Thanks for the review @alliomeria . Night

Issue #305, opened 2024-02-16
Basic IABookreader Search Controller misbehaves when a CWS Child uses it directly


DiegoPino:
# What?

Because our most basic Flavor driven Search Controller for OCR is unaware of IIIF Manifests (and we have now the Search API enabled endpoints that are really smart), when a CWS child has a IABookreader attached directly and e.g has a PDF, the Page used to return the results is swapped for the relative to the parent Sequence ID, making all results appear on the same page.

This controller was always meant to deal with the simplest use cases. All media of a normal ADO being displayed (e.g a single PDF) or each CWS child being a single page. 

The fix (very specific for this use case) is simple but won't resolve the issue of e.g a CWS IIIF Manifest with complex multi page per CWS child data being displayed at the top level Object, so really a definitive fix requires 
- IIIF Search API that we use for Mirador to be used *config option + JS * also on IABookreader
- Make also Mirador react to the same URI `#fragments` that book reader accepts so Mirador can be used as a direct swap replacement of IA Book reader

Once question I have for you @alliomeria is:
- If I detect that e.g the results are coming from multiple children and some are already PDFs (means book reader won't be able to handle the search) should I basically abort highlighting and return a message saying we can't search? Or have a way that IABookreader for those use cases basically cancels even the search at all before we expose it? I kinda feel giving people wrong highlights is worst that not even enabling them ... what are your feelings about that?

DiegoPino:
Resolved via https://github.com/esmero/strawberryfield/commit/4a8a450072f108234ad8dfdcf508d0c91e766904

PR #306, opened 2024-02-16
ISSUE-305: Don't use parent_top_sequence_id if the result Node is the same as the c…

DiegoPino:
…aller node (the node that shows the book reader)

See #305


DiegoPino:
@alliomeria won't merge this one yet. I know it works but only I have tested it ...

DiegoPino:
And I will merge now

DiegoPino:
@alliomeria this fix is provisional and deals with Objects that are children of a CWS but have Bookreader enabled directly on them instead of doing the normal thing to do. Show the viewer at the CWS side. The full / better fix in the future (but not on this module) is to allow IABookreader to use A) IIIF Content Search API OR store in the JS the actual image of each page as an index, that way we could return from search which images matched and at the IABookreader level discard results from images not being present (instead of the internal behavior of their JS which is just a page number)

PR #332, opened 2024-09-30
1.4.0: Sync to main

DiegoPino:
Syncs main with 1.4.0. @alliomeria this one is different bc 1.4.0 diverged from main a while ago. So I create a rebase 1.4.0 (named differently) The code is the same as 1.4.0, but with a different change history. Sorry and thanks


alliomeria:
Ok, understood. Thanks @DiegoPino.

Issue #333, opened 2024-10-02
JMESPATH splitting by ',' in Key name providers fails to detect commas that are inside an expression


DiegoPino:
# What?

I should have known better. But a complex JMESPATH expression e.g

```JMESPATH
[date_created_edtf][?"date_type" == 'date_range'][ join('/',[date_from,date_to])][]
````
Will be split in two because we blindly split by comma

The solution is to use regular expression for splitting.

In specific this little one

`$pattern = '/[,]+(?![^\[]*\]|[^\(]*\)|[^\{]*\})/';`

@patdunlavey you will need this to test your date ranges! Pull coming before you start your day I hope (testing in production now)
@alliomeria server indexing again!

Issue #224, opened 2022-06-29
Add more Processing Options to JMESPath based KeyName providers


DiegoPino:
# What?

New options need to be 
- Normalize Strings
  - Lower Case
  - Title Case  
  - Trim spaces
- Date Granularity Deduplication (like if a date is a single day, the start/end date will be the same but different hours. Indexing 2 dates eventually
- Date Range (Drupal way) to allow Native date Range to be indexed into Solr
 
@patdunlavey @aksm @alliomeria @karomabiles any other data/string manipulations we need before sending the JSON keys data into Solr?

patdunlavey:
For date range indexing in solr, I'm thinking two things would be needed:
1. In https://github.com/esmero/strawberryfield/blob/1.4.0/src/Plugin/StrawberryfieldKeyNameProvider/JmesPathNameProvider.php, add a checkbox to appear in the JmesPath Strawberry Field Key Name Provider configuration that would appear after the user clicks "Is Date?" that would read (something like) "Generate as Date Range (Check this option if  - AND ONLY IF - you plan to index this value as a date range in solr.)"
2. Here: https://github.com/esmero/strawberryfield/blob/1.4.0/src/Plugin/DataType/StrawberryValuesViaJmesPathFromJson.php#L170, after all the processes have generated a list of dates, check to see if the date_range option was checked in the configuration, and if so, winnow $values down to min and max, and then replace the two $values with an object with `:value` and `:end_value` (see [`search_api_solr\Plugin\search_api\processor\DateRange::preprocessIndexItems`](https://git.drupalcode.org/project/search_api_solr/-/blob/4.x/src/Plugin/search_api/processor/DateRange.php?ref_type=heads))

DiegoPino:
exactly. That's the key

patdunlavey:
[...]

PR #334, opened 2024-10-03
issues/333-jmespath-splitting ---> 1.4.0

patdunlavey:
Use preg_split with regex to explode comma-separated jmespaths into individual jmespaths because individual jmespaths may contain commas.

See https://github.com/esmero/strawberryfield/issues/333


patdunlavey:
See also https://github.com/esmero/strawberryfield/issues/224

DiegoPino:
@patdunlavey thanks so much for opening this pull, please see #335 (jinx). I used the same regular expression I wrote last night but added a few more fail safes in case it dies. Also includes the date RANGE! code ... in case you want to give it a try?

DiegoPino:
Also I don't remember this level of "JINX" before. Like our pulls came seconds apart! hahaha

PR #351, opened 2025-03-12
ISSUE-349 + 768 Vector

DiegoPino:
# What? 

Adds a helper method that counts Flavours at different hierarchies (Direct, Children, Grandchildren) using configurable OR/AND 

Also adds the 768 Vector, the ability to have sub-flavor ids (needed for ML when a single PDF generates multiple annotations/detections) and some small Release bug fixes when re-tracking Flavour data sources at the Search API index level (we had some WARNINGs there) + Typo corrections

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.1",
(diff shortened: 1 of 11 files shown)


DiegoPino:
You will have to trust me on this one

Issue #352, opened 2025-03-14
Files Ingested via the JSON API and composted are not allowed to be removed by the queue worker


DiegoPino:
# What?

Oh well. This seems to be on me

The composting action of any File that ends being a temp during the CRUD processes is triggered correctly because it works at an Event Subscriber Level (so how the File/ADO was ingested does not matter) but, the JSON API requires a File to be upload upfront into a special place/computed Field 
```PHP
 $fields['field_file_drop'] = BaseFieldDefinition::create('entity_reference')
    ->setName('field_file_drop')
    ->setLabel(t('Drop Files'))
    ->setComputed(TRUE)
    ->setRevisionable(FALSE)
    ->setTranslatable(FALSE)
    ->setCardinality(FieldStorageDefinitionInterface::CARDINALITY_UNLIMITED)
    ->setReadOnly(FALSE)
    ->setTargetEntityTypeId($entity_type->id())
    ->setSettings(
      [
        'target_type' => 'file',
        'file_directory' => 'sbf_tmp',
        'uri_scheme' => $schema,
      ]
    )
    ->setClass(StrawberryFieldFileComputedItemList::class)
    ->setDisplayConfigurable('view', FALSE)
    ->setDisplayConfigurable('form', FALSE);
```

This is needed because Drupal has a strange Bundle -> File Hierarchy in the JSON API (means files can't exist as Entities there without a Bundle endpoint that resolves it. 

Because of this, those files (if  $schema == private:// exists) will end in

private://sbf_tmp/

Once managed and copied to their final destination we issue a Compost Event/Queue entry for that original Location, but the Compost Worker ("the on me part") is very strict on not allowing other modules/external agents to delete files ...
So we define only these paths as Safe

https://github.com/esmero/strawberryfield/blob/e2f1a61f7de988172a206db4262d486e7e7d903f/src/Plugin/QueueWorker/CompostQueueWorker.php#L189-L194

When the worker hits those JSON API ingested files, it will ignore them and stop trying to delete them (we removed the re-tries bc some users had too many files we knew can not be delete (e.g source is at s3:// == the destination the same place )

Oh, well. I have to fix this

In the meantime, anything in that folder can always be deleted manually (give or take +5 min of lifetime)

DiegoPino:
[...]

Issue #350, opened 2025-03-11
Allow Hydroponics to Take over (or support/complement) Search API Indexing


DiegoPino:
# What?

As you all know AMI is super fast. WE have users going for 15K-20K a day ingesting. Search API's internal Cron based indexing can't handle the needs of having so much indexed.

But Hydroponics could use a few seconds of its cycle to run Indexing on the background without doing any harm and at least advance a few thousands while other core tasks do the same.

Easy-peasy ... but require some config and small changes.

PR #355, opened 2025-05-25
ISSUE-352 and ISSUE-350: Composer improvements/logging and new Search API Hydroponics indexer

DiegoPino:
See #352 #350 

This cleans up hydroponics, adds some more verbose logging, adds Search API Indexing capabilities to background processing also cleaning some basics on the composter.

There is also a small cleanup open the JSON PATCH action to be more verbose when running via AMI (but independent)


DiegoPino:
@alliomeria this is ready. But will merge tomorrow AM because I started fixing Cantaloupe on another thread.

Issue #362, opened 2025-10-28
D11 compatibility Changes for 1.6.0 (or 2.0.0)


DiegoPino:
Corresponding macro to https://github.com/esmero/format_strawberryfield/issues/544

We have some deep core deprecations so will move slowly testing step by step. Mostly function signatures but also methods that need to be called manually (Strawberryfield serializer was one tricky one)

DiegoPino:
Closed as resolved

PR #363, opened 2025-10-28
ISSUE-362: D11 Deprecations and Changes

DiegoPino:
See #362 and https://github.com/esmero/format_strawberryfield/issues/544

@alliomeria Note (same as before). This is against 2.0.0 but that might change. The version against 1.6.0 is slightly different (*Symfony/signatures + Config Form Constructors are different/at least so far)

Will open a few but will mark them all as "draft" to avoid any merges.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "strawberryfield/strawberryfield",
-    "description": "A strawberry field for Drupal 10/11",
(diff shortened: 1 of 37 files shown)


DiegoPino:
@alliomeria 1.6.0 has now (direct commit to 1.6.0) the same features, minus the deprecation fixes and function signatures this one brought. I have not tested yet, but on it. Thanks

Issue #369, opened 2026-03-08
Implement ADO specific Drupal tokens


DiegoPino:
Supersedes #179 

The most needed tokens are

`node:ado:type` --> Useful for site building and Path Aliases
`node:ado:collections_in_breadcrumb` --> Stats (as dimensions)
  --> this one could be also `node:ado:parents_by_type:{the type} to fetch only collections
`node:ado:collections_in_breadcrumb` --> Stats (as dimensions)
  --> this one could be also `node:ado:parents_by_rel:{the relationship} to fetch only ispartof, ismemberof, etc
`node:ado:breadcrumb` --> Stats ( as dimensions)
`node:ado:as_{filetype}` --> acts like the CORE file level tokens extracting only images, videos, audios, etc
`node:ado:sbf_key_name:{keyname}` --> re uses key name providers. One could use node:field:property instead  too (so not sure about this one)

PR #381, opened 2026-05-07
ISSUE-369 & ISSUE-380: D11. New parent path Tokens but as annotations

DiegoPino:
See #370  and #369 

Also brings #380  + a small thing (keep walking) with new small vectors (3 and 12 as dimension) needed in the future for color matching/comparison


DiegoPino:
Need to cherry pick the new vectors for 1.7.0. So won't merge JUST yet

DiegoPino:
Ok. Cherry picked directly into 1.7.0. Wearing a black release manager hat  today, and I am not to be bothered! (hahaha)

Issue #384, opened 2026-06-15
Add BWMetaEdit to Technical metadata extraction


DiegoPino:
# what?

Only acts on WAV files. New settings/binary custom compiled under `esmero/php-8.3-fpm:1.7.0-multiarch `, the output is XML so we JSON-i-fy it. MD5 checksum (generated + stored) are also validated under a subkey `MD5match` under "Technical". E.g for a WAV without stored checksum    `"MD5match": null`
```JSON
{
 "flv:bwfmetaedit": {
                "Technical": [
                    {
                        "FileSize": "46941194",
                        "DateCreated": "2026-06-15 19:37:49",
                        "DateModified": "2026-06-15 19:37:49",
                        "Format": "Wave",
                        "CodecID": "0001",
                        "Channels": "2",
                        "SampleRate": "48000",
                        "BitRate": "1536000",
                        "BitPerSample": "16",
                        "Duration": "00:04:04.484",
                        "bext": "No",
                        "INFO": "Yes",
                        "Cue": "No",
                        "XMP": "No",
                        "aXML": "No",
                        "iXML": "No",
                        "MD5Generated": "1C993A8FA1BFF0ABFC7322209AFD9F01",
                        "MD5match": null
                    }
                ],
                "Core": [
                    {
                        "IART": "Elsa Culler",
                        "INAM": "Ballade, Six Pieces Op. 118 No. 3"
                    }
                ]
            },
            "sequence": 1
        }

```

With stored MD5, and on the fly generated being the same

`"MD5match": true`

```JSON
{
 "flv:bwfmetaedit": {
                "Technical": [
                    {
                        "FileSize": "46941194",
                        "DateCreated": "2026-06-15 19:37:49",
                        "DateModified": "2026-06-15 19:37:49",
                        "Format": "Wave",
                        "CodecID": "0001",
                        "Channels": "2",
                        "SampleRate": "48000",
                        "BitRate": "1536000",
                        "BitPerSample": "16",
                        "Duration": "00:04:04.484",
                        "bext": "No",
                        "INFO": "Yes",
                        "Cue": "No",
                        "XMP": "No",
                        "aXML": "No",
                        "iXML": "No",
                        "MD5Generated": "1C993A8FA1BFF0ABFC7322209AFD9F01",
                        "MD5Stored": "1C993A8FA1BFF0ABFC7322209AFD9F01",
                        "MD5match": true
                    }
                ],
                "Core": [
                    {
[...]

PR #385, opened 2026-06-15
ISSUE-384: adds bwfmetaedit to tech metadata pipeline

DiegoPino:
Includes also BWF mimetype in case someone ingests .bwf files (which are WAV files)

See #384


alliomeria:
This looks great too @DiegoPino!

